PR #35, opened 2018-06-20
Fix namespace for TS, clarify and complete layout policy, layout_right, and layout_left

hcedwar:


P0009/P0009.bs
```diff
@@ -989,59 +998,58 @@ Table � — Layout mapping policy and layout mapping requirements
...
+  <td>*Requires:* `0 <= m(i...)` </td>
 </tr>
 <tr>
-  <td>`m(i...)`</td>
+  <td>`m.required_span_size()`</td>
   <td>`E::index_type`</td>
-  <!-- TODO Review this wording! -->
-  <td>*Returns:* The distance between the start and the element of the multi-index space denoted by `i...` in the codomain `span` of the `basic_mdspan` for which `*this` is a layout mapping. </td>
-  <td>*Requires:* `0 <= m(i...) < m.required_span_size()`</td>
+  <td>*Returns:* one plus the maximum value of `m(i...)`.</td>
```
crtrott: what if a layout mapping co-domain doesn't start at 0?

P0009/P0009.bs
```diff
@@ -989,59 +998,58 @@ Table � — Layout mapping policy and layout mapping requirements
...
-  <td>*Returns:* !!!TODO!!!</td>
+  <td>*Returns:* `true` if `m(i...) != m(j...)` for every `i... != j...` </td>
   <td></td>
 </tr>
 <tr>
-  <td>`M::is_always_contiguous()`</td>
+  <td>`m.is_contiguous()`</td>
   <td>`bool`</td>
-  <td>*Returns:* !!!TODO!!!</td>
+  <td>*Returns:* `true` if the set values defined by `m(i)...` is equal to `[0..m.required_span_size())` </td>
```
crtrott: Not quite sure that this is correct. Could we do something like" true if for every value j with 0<=j<required_span_size() there exists a i... so that m(i)... ==j?

P0009/P0009.bs
```diff
@@ -989,59 +998,58 @@ Table � — Layout mapping policy and layout mapping requirements
...
   <td></td>
 </tr>
 <tr>
-  <td>`M::is_always_strided()`</td>
+  <td>`m.is_strided()`</td>
   <td>`bool`</td>
-  <td>*Returns:* !!!TODO!!!</td>
+  <td>*Returns:* `true` if `m(j...) - m(i...) = s[r]` when all members of `j...` and `i...` are equal except for exactly one `r`*th* member such that `j[r] == i[r]+1`. `s[r]` is the *stride* of ordinate `r`.</td>
```
crtrott: Does this wording sufficiently restrict it to for a given r s[r] is independent of the chose i.. and j...

PR #36, opened 2018-06-20
P0009: Put "exposition only" comments in italics

mhoemmen:


dhollman:
I think it does. You just have to use html code blocks directly instead of the backticks. Bikeshed lets you mix html and markdown.

mhoemmen:
@dhollman Let me see if I can figure that out....

mhoemmen:
The `<xmp>` tag (otherwise easier to use, since it doesn't require escaping less-thans) has the same issue:
```
<xmp highlight="c++">
std::vector<int> foo (42); // *exposition only*
</xmp>
```

mhoemmen:
I tried replacing the asterisks with `<em> ... </em>` or `<i> ... </i>`; no luck.

mhoemmen:
Oh wait, I got it!  This works:
```
<pre highlight="c++">
std::vector&lt;int> foo (42); // <i>foo</i>
</pre>
```

mhoemmen:
Yup, it has to be `<pre>`, not `<xmp>`.  This will take a while to fix....

PR #43, opened 2018-06-21
P0009: Italicize "exposition only" comments

mhoemmen:
This unfortunately requires a `<pre>` block instead of triple-backticks or `<xmp>`, hence the removal of left angle brackets.


mhoemmen:
YAY :D

mhoemmen:
FYI, I looked at JF's papers in Bikeshed and they all use `<pre highlight="c++"> ... </pre>` blocks for code.

PR #55, opened 2018-07-07
Introduce beginnings of prototype/reference implementation, adjust wording

hcedwar:
Introduce beginnings of prototype / reference implementation based upon updating prior prototypes to new wording.  
Adjust wording based upon prototyping experience.
Non-trivial change to the `Accessor` to allow a non-template `Accessor` class to denote access properties independent of the element type.

### DIFF
--- a/P0009/P0009.bs
+++ b/P0009/P0009.bs
@@ -1869,7 +1869,7 @@ static constexpr reference deref( accessor_type acc , size_t i ) noexcept ;
 * *Returns:* `acc[i]` <br/>
 
 ```c++
-static constexpr pointer deref( accessor_type acc , size_t i ) noexcept ;
+static constexpr pointer decay( accessor_type acc , size_t i ) noexcept ;
 ```
 
 * *Returns:* `acc` <br/>
--- a/P0009/P0009.bs
+++ b/P0009/P0009.bs
@@ -1786,6 +1786,12 @@ Table �: Accessor requirements
   <td></td>
   <td>*Requires:* `p` is a pointer to a contiguous set of objects.</td>
 </tr>
+<tr>
+  <td>`a = A(p);`</td>
+  <td></td>
+  <td>*Requires:* `p` is a pointer to a contiguous set of objects.</td>
+</tr>
+<tr>
 </table>
 
 <!--
@@ -1930,8 +1936,8 @@ public:
   static constexpr index_type required_span_size(const array&lt;IndexType, N>& dynamic_extents) noexcept;
 
 private:
-  mapping_type map_; // <i>exposition only</i>
-  pointer ptr_; // <i>exposition only</i>
+  mapping_type  map_; // <i>exposition only</i>
+  accessor_type acc_; // <i>exposition only</i>
 };
 
 }}}
@@ -1957,7 +1963,7 @@ constexpr basic_mdspan() noexcept = default;
 <!-- TODO check for periods, commas, and "and" here. -->
 <!-- TODO: decide if we want to include invocation of `span()` in the postconditions. An accessor may want to make this undefined behavior. -->
 * *Effects:*
-    * zero-initializes `ptr_`
+    * zero-initializes `acc_`
     * value-initializes `map_`
 * *Postconditions:* 
     + `size()==0`
@@ -1971,7 +1977,7 @@ constexpr basic_mdspan(const basic_mdspan& other) noexcept = default;
 ```
 
 * *Effects:*
-    + initializes `ptr_` with `other.ptr_`
+    + initializes `acc_` with `other.acc_`
     + initializes `map_` with `other.map_`
 * *Postconditions:*
     + `size()==other.size()`
@@ -1985,7 +1991,7 @@ constexpr basic_mdspan(basic_mdspan&& other) noexcept;
 ```
 
 * *Effects:*
-    + initializes `ptr_` with `move(other.ptr_)`
+    + initializes `acc_` with `move(other.acc_)`
     + initializes `map_` with `move(other.map_)`
 
 <br/>
@@ -1997,7 +2003,7 @@ template<class... IndexType>
 
 * *Requires:* `[ptr, ptr+required_span_size(dynamic_extents...))` shall be a valid range.
 * *Effects:*
-    + initializes `ptr_` with `ptr`
+    + initializes `acc_` with `ptr`
     + initializes `map_` with `extents_type(dynamic_extents...)`
 * *Postconditions:* 
     + `extents()==extents_type(dynamic_extents...)`
@@ -2017,7 +2023,7 @@ template<class... IndexType>
 
 * *Requires:* `sp.size()==required_span_size(dynamic_extents...)`
 * *Effects:*
-    + initializes `ptr_` with `sp.data()`
+    + initializes `acc_` with `sp.data()`
     + value-initializes `map_`
 * *Postconditions:* 
     + `extents()==extents_type(dynamic_extents...)`
@@ -2066,7 +2072,7 @@ constexpr basic_mdspan(pointer p, const mapping& m);
 
 * *Requires:* `[ptr, ptr+m.required_span_size())` shall be a valid range.
 * *Effects:*
-    + initializes `ptr_` with `p`
+    + initializes `acc_` with `p`
     + initializes `map_` with `m`
 * *Postconditions:* 
     + `extents()==m.extents()`
@@ -2092,7 +2098,7 @@ constexpr basic_mdspan(pointer p, const mapping& m, const accessor& a);
 
 * *Requires:* `[ptr, ptr+m.required_span_size())` shall be a valid range.
 * *Effects:*
-    + initializes `ptr_` with `p`
+    + initializes `acc_` with `p`
     + initializes `map_` with `m`
 * *Postconditions:* 
     + `extents()==m.extents()`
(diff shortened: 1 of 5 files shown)


P0009/P0009.bs
```diff
@@ -1805,28 +1796,16 @@ namespace std {
...
 };
 
 }}}
 ```
 
-<br/>
-<b>26.7.�.2.1 `accessor_basic` operations [mdspan.accessor.basic.ops]</b>
-
-```c++
-constexpr reference operator()(pointer p, ptrdiff_t i) const noexcept;
```
crtrott: We think it is okish to delete that operator, but there are drawbacks and benefits:
- benefit: common case, pointer is just T* has now one less inlined function.
- drawback: requires pointer to be more often a wrapper object which contains all the stuff the accessor would contain, for example for the atomic accessor, pointer is not anymore T*. It needs to be some wrapper class which is convertible to T* and which has an [] operator.
- drawback: pointer must be templated on reference, since its operator [] must return that. 
- drawback: no clear separation of access and storage.
     - e.g. if I want to use some intrinsic to access the data in previous solution that intrinsic is in the operator of the accessor, now I write a little class containing a pointer and a [] operator, and the implementation of the [] operator contains the intrinsic

It seems this makes stuff more complex down the road. My main issue is that pointer[int] must return a reference, which means pointer and reference are not independent any more.

P0009/P0009.bs
```diff
@@ -1763,24 +1759,19 @@ Table �: Accessor requirements
...
   <td></td>
-  <td>*Requires:* `A::pointer` shall meet the requirements of random access iterator (**[random.access.iterators]**), and `iterator_traits<A::pointer>::value_type` shall be exactly `A::value_type`.</td>
+  <td>*Requires:* `A::reference<T>` shall be convertable to `T` and, if `!is_const_v<T>`, assignable from `T`. *[Note:* This may be `T&` or a proxy that operates like `T&`. *--end note]*</td>
 </tr>
 <tr>
-  <td>`A::reference`</td>
+  <td>`A::pointer<T>`</td>
   <td></td>
-  <td>*Requires:* `iterator_traits<A::pointer>::reference` shall be convertible to `A::reference` </td>
+  <td>*Requires:* `A::pointer<T>` shall be `DefaultConstructible` and `CopyAssignable`. *[Note:* This may be `T*` or a proxy that operates like `T*`. *--end note]*</td>
```
crtrott: This doesn't seem to be enough in terms of what pointer is. This for example doesn't define that operator[] (ptrdiff_t) exists or how that behaves. It doesn't define that operator* () exists or what it does. And we need arithmetic on those to allow subspan. If we define all of those than we are probably getting really close to random access iterator. Is there anything specific in random access iterator we don't want or shouldn't require for pointer?

P0009/P0009.bs
```diff
@@ -1001,13 +1003,9 @@ Table � — Layout mapping policy and layout mapping requirements
...
 <tr>
-  <td>`m.get_extents()`</td>
-  <td>`E`</td>
-  <td>
-  
-  *Returns:* `e`.
-  
-  </td>
+  <td>`m.extents()`</td>
+  <td>convertable to `E`</td>
```
crtrott: Is there a good reason to not return an actual Extent? We'd rather have the more specific one. vector<bool> is a good example where this is a problem, since it returns something convertible to bool and not a bool. Which means that for example doing auto and than using the variable in a templated function is an issue.

P0009/P0009.bs
```diff
@@ -2364,32 +2316,51 @@ constexpr mapping get_mapping() const noexcept;
...
 ```
 
-* *Returns:* `get_mapping().stride(r)`
+* *Returns:* `mapping().stride(r)`
+
+<br/>
+
+```c++
+template<class... IndexType>
+static constexpr index_type required_span_size(IndexType... dynamic_extents) noexcept;
```
crtrott: We discussed this and were not in favor of replicating too much stuff in mdspan which exists in mapping. But it wasn't a super strong preference. If you like it to be in here that is fine.

P0009/P0009.bs
```diff
@@ -2267,34 +2232,21 @@ static constexpr index_type static_extent(size_t r) noexcept;
...
 constexpr index_type extent(size_t r) const noexcept;
 ```
 
-* *Returns:* `get_extents().extent(r)`.
+* *Returns:* `extents().extent(r)`.
 
 <br/>
 
-```c++
-constexpr index_type size() const noexcept;
```
crtrott: We believe we need size and unique_size. One example is I want to pack a strided view into an MPI buffer. I need to know how large to make that buffer, and its not required_span_size. And if we add size we probably also need a function which returns something discounting all the replicated entries in the mapping.

P0009/P0009.bs
```diff
@@ -1805,28 +1796,16 @@ namespace std {
 namespace experimental {
 namespace fundamentals_v3 {
 
-template<class T>
 struct accessor_basic {
```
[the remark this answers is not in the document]
hcedwar: The gain is to be able to define and compose access properties independently of the type to which they are applied; e.g., atomic, streaming, etc.  Cumbersome to compose access properties if the ElementType must be carried along during composition, as apposed to being applied after the fact.  Consider a "rebind" or "apply to" interface:
  `auto ax = apply_property( x , access_atomic );`
where
  `inline constexpr const access_atomic_t access_atomic;`

P0009/P0009.bs
```diff
@@ -1112,7 +1110,7 @@ struct layout_left {
       constexpr mapping& operator=(const mapping&lt;OtherExtents>& other);
 
     // [mdspan.layout.left.ops], layout_left::mapping operations
-    Extents get_extents() const noexcept;
```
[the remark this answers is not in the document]
hcedwar: at least weaken to *convertable to Extents* so an implementation is not forced to have the penalty of constructing an extent every time the function is called.

P0009/P0009.bs
hcedwar: Yep - not independent.  The *pointer* and *reference* should have a strong relationship.  The *pointer* is carried around as a member of mdspan and is handle to storage, the *reference* is an associated handle to accessing an element of that storage.  Should just be able to copy the storage handle and everything works - no extra wrappers.
See the prototype where `template<class T> using pointer = T *;` for the accessor_basic.

P0009/P0009.bs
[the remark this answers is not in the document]
dhollman: Actually, the code you gave originally:
```
auto ax = apply_property( x , access_atomic );
```
can just work as is, assuming an implementation of `apply_property` that deduces the type of `x` and acts on some sentinel `access_atomic`, especially if we add an optional `rebind` member template, like on allocator. I wouldn't be opposed to that, because that's something I think we could reasonably allow at the concept level without losing the ability to allow accessors to restrict the types they act on and partially specialize.

P0009/P0009.bs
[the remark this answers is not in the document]
dhollman: From the CppCoreGuidelines:

> What is “cheap to copy” depends on the machine architecture, but two or three words (doubles, pointers, references) are usually best passed by value. 

Most `extents` will be smaller, so our concrete types should return by value IMHO. But I agree, the concept should not have this constraint.

P0009/P0009.bs
hcedwar: Operational spec required `p[i]` to return a reference to the offset - so that is still there.
An iterator also requires `*p`, `++p`, and `p++` - which should not be required by operational spec.

P0009/P0009.bs
hcedwar: Convertable allows return by value or return by constant reference, which allows an implementation to avoid the overhead of being required to construct and return a copy.

P0009/P0009.bs
dhollman: It's not just about replicating too much stuff. This won't work for non-unique, stateful mappings, so the operational semantics would have to somehow exclude those mappings, and the mapping would have to somehow advertise that its statefulness affects its required span size. I'm okay with putting it back in if those problems are solved, but this pull request doesn't solve those problems AFAICT.

hcedwar: The `mapping` version requires construction of a mapping object and then function call on that object, a syntactically long two-step operation .  I believe this will be a frequent pattern and the shortcut is a desirable convenience.  The other single-step operations could be removed.

P0009/P0009.bs
hcedwar: Pull them from the mapping :-)

dhollman: The reasoning for having `size()` and `unique_size` here is for consistency with container-like types (this is going into the containers subclause of the standard after all) in the case of `size()`, and for clarity as to which size the `size()` method refers to in the case of `unique_size`.

P0009/P0009.bs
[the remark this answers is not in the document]
crtrott: Doesn't that mean as in the case below that we end up having problems with stuff like auto on the return type of extents? Also convertable to Extents doesn't solve the issue of creating a copy. I either return a reference to something that has to exist in mapping (in which case changes to mapping would be observable in the returned reference I got via `auto& e=m.extents();`  , or I return a copy in which case I don't need to prescribe what is in mapping.
So two options:
  * return const Extents& and mapping has now a real member of Extents not just exposition only, but it fixes the "I need to create a copy" issue. 
  * leave it as is. 

Thoughts on what is preferable?

P0009/P0009.bs
dhollman: The point about `*p`, `++p`, and `p++` not needing to exist is interesting. Do you envision any cases where this would be necessary? You're losing a lot of the power you get by separating storage from access, and it's not clear how you would form the codomain span with something that doesn't meet the requirements of `RandomAccessIterator`.

crtrott:
Closing this now. With the just merged commit of a reference implementation we should have gotten everything transferred from here.

PR #63, opened 2018-08-25
P0009: Adding subspan changes adopted from PR55

crtrott:


P0009/P0009.bs
```diff
@@ -2405,43 +2405,50 @@ namespace fundamentals_v3 {
...
 
 }}}
 ```
 
+
 `subspan` creates a `basic_mdspan` that is a view on a (potentially trivial) subset of another `basic_mdspan`.
 The SliceSpecifier parameters indicate the subset that the return value references.
-Let slices[r] denote the rth value in the parameter pack slices.
-The second template parameter of the return type is E, a specialization of `extents`.
+The second template parameter of the return type, `E`, is a specialization of `extents`.
```
dsunder: Strike "the second template parameter of the return type".
i.e.
`E` is a specialization of `extents`.

crtrott:
Dan said that we should do a pass through this together next week and then do those changes, since they need to be done all over the paper.

PR #64, opened 2018-08-26
Reference impl

crtrott:


crtrott:
This is based on Carters Pull request, but heavily reworked and fixed to on one hand be in line with the current paper draft, and on the other add missing things like subspan etc. Note that we need to do more cleanup. This still contains stuff which is not described in the paper and should be removed. 

There is a testing directory with a simple build script. It uses gtest. Gonna write a proper Makefile later.

crtrott:
Just gonna merge this now, since I doubt anybody is going to read the 50k+ lines in detail ...

PR #67, opened 2018-09-07
P0009 reference extents

hcedwar:
Static and non-static extents with r >= rank() are specified to be 1


crtrott:
Jup this is in the wording.

PR #74, opened 2018-10-01
P0009r8: Fix various issues including #73

mhoemmen:
@crtrott @dsunder @dhollman

Issue #76, opened 2018-10-02
P0009: Accessor policy: What "additional restrictions" may the policy impose?


mhoemmen:
@crtrott @dsunder @dhollman

Issue #75, opened 2018-10-02
P0009: What are the requirements on `reference in accessor policy?


mhoemmen:
@crtrott @dsunder @dhollman

dhollman:
I think in general, all we can hope for is "convertible to `ElementType`".  However, for this to be useful in any remotely generic context, we'll eventually have to add special requirements for when `ElementType` is numeric.  We can always add these after we reach TS, though.

mhoemmen:
@dhollman I do think we need _something_ here, else LWG may kick this back even for a TS.  How about this?

  - If `is_const_v<element_type>`, then `is_convertible_v<reference_type, element_type>`;
  - Otherwise, `is_convertible_v<reference_type, element_type> && is_assignable_v<reference_type, element_type>`.

We also want something about how the accessor policy may impose additional constraints.

dhollman:
I think that's probably what we'd want to put in the paper for now. It doesn't preclude any of the reference types we've been imagining, and it allows us to add more later for special cases like numeric types.

dhollman:
That sounds fine.  I think that was my only substantive change to #77; IIRC the rest were editorial.

mhoemmen:
Just had a phone discussion with @crtrott .  He convinced me that `reference` need not necessarily have anything to do with `element_type`, in terms of Constraints.  For example, the accessor policy defines how to turn a `pointer` into `element_type*` (via `decay`); this makes the `mdspan` method that returns a `span` well defined.

I'm guessing that `reference` should be _Cpp17CopyConstructible_ or _Cpp17CopyAssignable_`. 
 Otherwise, it doesn't make sense to return `reference` by value.  However, perhaps we don't need to _say_ that.

We could perhaps add a paragraph of Notes before the accessor policy definition that explains the intent and gives an example (other than `basic_accessor`).

dhollman:
> I'm guessing that reference should be Cpp17CopyConstructible or Cpp17CopyAssignable.
Otherwise, it doesn't make sense to return reference by value.

It could be move-only. In fact, with mandatory RVO in C++17, it doesn't even need to have a move constructor.

But the bigger issue for me is that its impossible to do anything in code that is generic over the `AccessorPolicy` type unless you have some semantic guarantees about the reference.  Remember, the set of operations we require on `AccessorPolicy::reference` are the *only* operations you're technically allowed to do in code that is generic over `AccessorPolicy`.  @crtrott how do you intend to get around this?

[...]

PR #77, opened 2018-10-02
P0009r8: Math-ify mdspan.subspan; clean up accessor policy; fix #71

### LINKED ISSUE #71: P0009: required_span_size inconsistent definition?
@crtrott @dsunder 

The specific layouts define `required_span_size()` as the product of `extents().extent(r)` for all `r` where 0 <= `r` < `extents().rank()`, but the requirements table defines it as 1 plus the maximum value of `m(i...)`.  These definitions contradict for a rank-0 `mdspan`, or for a degenerate`mdspan`.  "Degenerate" means that the `mdspan` has nonzero rank, but one or more dimensions is zero.

Kokkos just defines the `span()` as the product of the dimensions.  Thus, I propose the following fix for the definition of `required_span_size` in the table:
```
* *Returns:*
    * If the codomain has size zero, then zero;
    * Otherwise, 1 plus the maximum value of `m(i...)`.
```

mhoemmen:
- Math-ify mdspan.subspan, as discussed with @crtrott today in #74 comments
- Clean up accessor policy, per discussion with @crtrott today
- Fix #71, at least partly

"At least partly" means that we may still need to look at the definitions of `required_span_size()` for specific layouts.

### DIFF
--- a/P0009/P0009.bs
+++ b/P0009/P0009.bs
@@ -2459,19 +2459,31 @@ and is implementation defined.
 The `SliceSpecifier` parameters indicate the subset that the return value references.
 <br/>
 
-Define `slices_tuple` as `auto slices_tuple = forward_as_tuple(slices...);`. <br/>
-Define `first[r]` as follows: <br/>
-    * If `slices_tuple.get<r>()` is convertible to `ptrdiff_t`, then `first[r]` equals `slices_tuple.get<r>()`. <br/>
-    * Otherwise, if `slices_tuple.get<r>()` is convertible to `pair<ptrdiff_t,ptrdiff_t>`,
-      then `first[r]` equals `pair<ptrdiff_t,ptrdiff_t>(slices_tuple.get<r>()).first`. <br/>
-    * Otherwise, if `slices_tuple.get<r>()` is convertible to `all_type`, then `first[r]` equals `0`. <br/>
-Define `last[r]` as follows: <br/>
-    * If `slices_tuple.get<r>()` is convertible to `ptrdiff_t`,
-      then `last[r]` equals `slices_tuple.get<r>()+1`. <br/>
-    * Otherwise, if `slices_tuple.get<r>()` is convertible to `pair<ptrdiff_t,ptrdiff_t>`,
-      then `last[r]` equals `pair<ptrdiff_t,ptrdiff_t>(slices_tuple.get<r>()).second`. <br/>
-    * Otherwise, if `slices_tuple.get<r>()` is convertible to `all_type`,
-      then `last[r]` equals `src.extent(r)`. <br/>
+<pre highlight="c++">
+  auto slices_tuple = forward_as_tuple(slices...);
+  array<ptrdiff_t, sizeof...(slices)> first;
+  array<ptrdiff_t, sizeof...(slices)> last;
+  for(size_t r=0; r&lt;sizeof...(slices); r++) {
+    if constexpr (is_convertible_v<decltype(slices_tuple.get<r>()), ptrdiff_t>) {
+      first[r] = slices_tuple.get<r>();
+      last[r] = first[r] + 1;
+    }
+    else if constexpr (is_convertible_v<decltype(slices_tuple.get<r>()), pair<ptrdiff_t,ptrdiff_t>>) {
+      const pair<ptrdiff_t, ptrdiff_t> slices_r = slices_tuple.get<r>();
+      first[r] = slices_r.first;
+      last[r] = slices_r.second;      
+    }
+    else if constexpr (is_convertible_v<slices_tuple.get<r>(), all_type>) {
+      first[r] = 0;
+      last[r] = src.extent(r);
+    }
+    else {
+      // <i>emit an implementation-defined diagnostic</i>
+      static_assert(false);
+    }
+  }
+</pre>
+
 Define `R[K]` as follows: <br/>
 <pre highlight="c++">
   size_t k = 0;
--- a/P0009/P0009.bs
+++ b/P0009/P0009.bs
@@ -1108,12 +1108,12 @@ Table � — Layout mapping policy and layout mapping requirements
 <tr>
   <td>`m.is_strided()`</td>
   <td>`bool`</td>
-  <td>*Returns:* `true` if there exists an integer <math>s<sub>r</sub></math>
+  <td>*Returns:* `true` if there exists an integer <math>&sigma;<sub>r</sub></math>
        such that, for all `j...` and `i...` in `e`,
        if all members of `j...` and `i...` are equal
        except for exactly one r-th member where
        <math>j<sub>r</sub></math> equals <math>i<sub>r</sub></math> + 1, 
-       then `m(j...) - m(i...)` equals <math>s<sub>r</sub></math>.</td>
+       then `m(j...) - m(i...)` equals <math>&sigma;<sub>r</sub></math>.</td>
   <td></td>
 </tr>
 <tr>
@@ -1137,8 +1137,8 @@ Table � — Layout mapping policy and layout mapping requirements
 <tr>
   <td>`m.stride(r)`</td>
   <td>`E::index_type`</td>
-  <td>*Returns:* <math>s<sub>r</sub></math>, as defined for `m.is_strided()` above.
-      We say that <math>s<sub>r</sub></math> is the *stride* of ordinate `r`. </td>
+  <td>*Returns:* <math>&sigma;<sub>r</sub></math>, as defined for `m.is_strided()` above.
+      We say that <math>&sigma;<sub>r</sub></math> is the *stride* of ordinate `r`. </td>
   <td>*Requires:* `m.is_strided()` is `true`. </td>
 </tr>
 </table>
--- a/P0009/P0009.bs
+++ b/P0009/P0009.bs
@@ -1903,7 +1903,7 @@ Table �: Accessor policy requirements
 <tr>
   <td>`A::offset_policy`</td>
   <td></td>
-  <td>Accessor policy which can access a pointer that has been return by `a.offset(p,i)`.</td>
+  <td>Accessor policy which can access a pointer that has been returned by `a.offset(p,i)`.</td>
 </tr>
 <tr>
   <td>`a.decay(p)`</td>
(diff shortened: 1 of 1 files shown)


crtrott:
Looks pretty good to me: One thing: do we need the `only` in `a handle to a contiguous set of elements of type element_type, accessible only through the policy’s access function;`?

Also does the `other than slices...,` serves a particular purposes in `If i... is a pack of integers other than slices..., then denote the value of the k -th member of i... by i k.` and do we need the same sentence for `j...`?

crtrott:
And thanks for all your hard work, in particular since I made you redo a bit of it .... !!

mhoemmen:
@crtrott wrote:
> And thanks for all your hard work, in particular since I made you redo a bit of it .... !!

Awww, don't worry, it wasn't so bad :-) Thanks for taking a look tonight!  I'll do the following:

  1. Get rid of "only" in the sentence you mentioned
  2. Change the "other than `slices...`" bit: "Denote the value of the k-th member of `i...` by `<math>i<sub>k</sub></math>`, and denote the value of k-th member of `j...` by <math>j<sub>k</sub></math>`."

mhoemmen:
@crtrott btw there are some issues that I opened, that I think still need to be addressed: #75, #76, and #78.  I favor separating those fixes into separate pull requests.

mhoemmen:
@crtrott I just pushed a commit to this PR addressing your suggestions above, along with a `<math>...</math>` syntax error and some minor formatting issues.

P0009/P0009.bs
```diff
@@ -607,36 +607,42 @@ Description
...
-The layout mapping of a multi-index is intended to be an O(1) `constexpr`
-operation that is trivially inlined and optimized. Note that Fortran
-compilers' optimizations include loop invariant code motion, including
-partial evaluation of multi-index layout mappings when indices are
-loop-invariant.
+We intend the layout mapping of a multi-index to be a constant-time
+`constexpr` operation that is trivially inlined and optimized.
+Compiler vendors may apply optimizations such as loop-invariant code
+motion, including partial evaluation of multi-index layout mappings
+when indices are loop invariant.
```
dhollman: (minor note) "loop-invariant" was correct in the first place, since invariant is an adjective rather than a verb here, and "loop" doesn't have the traditional form of an adverb.  (Just a grammar nit-pick :-P)

P0009/P0009.bs
```diff
@@ -781,10 +813,10 @@ Y8b.     .d8""8b. Y88b. Y8b.     888  888 Y88b.       X88
 
 <b>26.7.�.1 Overview [mdspan.extents.overview]</b>
 
-1. An `extents` object defines a *multidimensional index space* which is the Cartesian product of integers extents `[0..N0) * [0..N1) *`....
-2. The *dynamic extents* of an `extents` object correspond to the `StaticExtents` template parameters that are equal to `dynamic_extent`.  Let *DynamicRank[i]* denote the index of the *i*th such extent in the `StaticExtents` template parameter pack, and let *DynamicIndex[r]* indicate the number of such extents in the first *r* entries of the `StaticExtents` parameter pack
+1. An `extents` object defines a multidimensional index space.
```
dhollman: We've included the definition of "multidimensional index space" in the introduction but not in the wording. Doesn't it also need to be in the wording?

P0009/P0009.bs
```diff
@@ -978,26 +1034,35 @@ template<ptrdiff_t... LHS, ptrdiff_t... RHS>
...
+    * `e` denotes an object of type `E` defining a multidimensional index space.
     * `r` is a value of an integral type such that 0 <= `r` < `e.rank()`.
-    * `i...` and `j...` are packs of an integer type denoting values in the multi-index space `e`, the `r`*th member of packs `i...` and `j...` are denoted by `i[r]` and `j[r]`, and `sizeof...(i)==E::rank()`, 0 <= `i[r]` < `e.extent(r)`, `sizeof...(j)==E::rank()`, and 0 <= `j[r]` < `e.extent(r)`.
+    * If `k...` is a pack of an integer type, we say that "`k...` is in `e`" 
+      if `k...` denotes a multi-index in `e`.
+      Let the packs of integer type `i...` and `j...` be in `e`.
+      Denote the <math>r</math>-th member of packs `i...` and `j...` by 
+      <math>i<sub>r</sub></math> and <math>j<sub>r</sub></math>, respectively. 
+      Let `r` equal <math>r</math>.
+      `sizeof...(i)==E::rank()` is `true`,
```
dhollman: Not sure why we say "is `true`" here, but if that's an LWG thing it's fine.

P0009/P0009.bs
```diff
@@ -1018,56 +1083,63 @@ Table � — Layout mapping policy and layout mapping requirements
...
 <tr>
   <td>`m.is_strided()`</td>
   <td>`bool`</td>
-  <td>*Returns:* `true` if `m(j...) - m(i...) = s[r]` when all members of `j...` and `i...` are equal except for exactly one `r`*th* member such that `j[r]==i[r]+1`. `s[r]` is the *stride* of ordinate `r`.</td>
+  <td>*Returns:* `true` if there exists an integer <math>&sigma;<sub>r</sub></math>
+       such that, for all `j...` and `i...` in `e`,
+       if all members of `j...` and `i...` are equal
+       except for exactly one r-th member where
+       <math>j<sub>r</sub></math> equals <math>i<sub>r</sub></math> + 1, 
+       then `m(j...) - m(i...)` equals <math>&sigma;<sub>r</sub></math>.</td>
```
dhollman: Does this also need to say `sigma_r > 1`?

P0009/P0009.bs
```diff
@@ -1207,8 +1279,8 @@ constexpr mapping(const mapping<OtherExtents>& other);
 
 * *Requires:* `other.extents()` meets the requirements for use in the initialization of `extents_`.
 * *Effects:* Initializes `extents_` with `other.extents()`.
-* *Postconditions:* `extents()==other.extents_`.
-* *Throws:* nothing.
+* *Ensures:* `extents()==other.extents_` is `true`.
```
dhollman: should this be `other.extents()` for consistency with the copy constructor?

P0009/P0009.bs
```diff
@@ -1618,93 +1681,131 @@ struct layout_stride {
...
 template<class OtherExtents>
   constexpr mapping(const mapping<OtherExtents>& other);
 ```
-* Requires: other.extents() meets the requirements for use in the initialization of extents_.
-* Effects: Initializes `extents_` with `other.extents()` and initializes `strides_` with `other.strides()`.
-* Postconditions: `extents()==other.extents()` and `strides()==other.strides()`.
-* Throws: nothing.
+* *Requires:* `other.extents()` meets the requirements for use in the initialization of `extents_`.
+* *Effects:* Initializes `extents_` with `other.extents()`, and initializes `strides_` with `other.strides()`.
+* *Ensures:* `extents()==other.extents() && strides()==other.strides()` is `true`.
```
dhollman: Should this be two lines for consistency?

P0009/P0009.bs
```diff
@@ -1807,12 +1913,12 @@ Table �: Accessor policy requirements
...
   <td>`a.access(p,i)`</td>
   <td>`A::reference`</td>
-  <td>*Returns:* An object which provides access to the `i`*th* element in the contiguous set.</td>
+  <td>*Returns:* An object which provides access to the `i`-th element in the contiguous set of elements. </td>
 </tr>
 <tr>
   <td>`a.offset(p,i)`</td>
   <td>`A::offset_policy::pointer`</td>
-  <td>*Returns:* A pointer for the contiguous span of elements offset by `i`.  *Requires:* `A::decay(p)+i == A::offset_policy::decay(A::offset(p,i))`.</td>
+  <td>*Returns:* A pointer to the contiguous span of elements offset by `i`.  *Requires:* `A::decay(p)+i` equals `A::offset_policy::decay(A::offset(p,i))`.</td>
```
dhollman: Should there be a line break here?

P0009/P0009.bs
```diff
@@ -2147,20 +2250,21 @@ template<class OtherElementType, class OtherExtents, class OtherLayoutPolicy, cl
   constexpr basic_mdspan& operator=(const basic_mdspan<OtherElementType, OtherExtents, OtherLayoutPolicy, OtherAccessor>& other);
 ```
 
+<!-- NOTE is_assignable_v<T,U> means T is assignable from U -->
+* *Constraints:*
+    + `is_assignable_v<ElementType(*)[], OtherElementType(*)[]>` is `true`, <!-- lifted directly from span wording -->
```
dhollman: Isn't the accessor pointer type assignment sufficient (below)? The above wording is only in `span` because it lacks the `Accessor` to constrain the pointer type, which could come from an allocator to a container that uses fancy pointers.  I don't think we have that problem.

P0009/P0009.bs
mhoemmen: @dhollman This is an LWG thing.  Boolean code-font statements are just statements that could evaluate to `false` or `true`.

mhoemmen: It's different when something Returns a `bool`; then it just returns the value of the code-font `bool` statement.

P0009/P0009.bs
mhoemmen: `layout_left``mdspan` are strided (`is_strided()` is `true`).  Consider the left-most index.  There, `m(j...) - m(i...)` equals 1.  Thus, <math>sigma<sub>r</sub></math> may be 1 in this case.

Whether <math>sigma<sub>r</sub></math> may be zero relates to the issue of degenerate mdspan, #78.

P0009/P0009.bs
mhoemmen: @dhollman I agree; this must be a left-over from the now deleted mdspan(span) constructor.

P0009/P0009.bs
```diff
@@ -607,36 +607,42 @@ Description
...
 
 **Optimized Implementation of Layout Mapping**
 
-The layout mapping of a multi-index is intended to be an O(1) `constexpr`
-operation that is trivially inlined and optimized. Note that Fortran
-compilers' optimizations include loop invariant code motion, including
-partial evaluation of multi-index layout mappings when indices are
-loop-invariant.
+We intend the layout mapping of a multi-index to be a constant-time
+`constexpr` operation that is trivially inlined and optimized.
```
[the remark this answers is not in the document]
mhoemmen: Fixed in the comments I just pushed.

P0009/P0009.bs
mhoemmen: Fixed in the comments I just pushed.

mhoemmen:
Thanks y'all! :D

PR #85, opened 2018-11-20
P1370: Add first draft of numeric_limits experiences paper

mhoemmen:
@dalg24 This is the original version of the paper that I sent you.  I have not yet made revisions.  Thanks!


mhoemmen:
@dalg24 , I just pushed one more commit with Sandia's funding statement and disclaimer.

crtrott:
is this still relevant?

dalg24:
This paper was a response to http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0437r1.pdf

Walter proposed wording for the numeric traits with the changes we suggested
http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2020/p1841r1.pdf

I think we should merge as is

mhoemmen:
LEWG reviewed some other paper recently that relates to this, if I remember right.

PR #99, opened 2019-06-23
P1673 & P1674 from pre-Cologne 2019 submission

mhoemmen:
Already submitted, but we should merge them into the public repo.  @crtrott @dhollman


p-groarke:
@mhoemmen Hi Mark, there is a copy-paste error in `accessor_conjugate::access`. This class doesn't have a `scale_factor` ;)

```c++
template<class Accessor, class T>
class accessor_conjugate {
public:
  using element_type  = Accessor::element_type;
  using pointer       = Accessor::pointer;
  using reference     = Accessor::reference;
  using offset_policy = Accessor::offset_policy;

  accessor_conjugate(Accessor a) : acc(a) {}

  reference access(pointer p, ptrdiff_t i) const noexcept {
    return reference(acc.access(p,i),scale_factor); // no scale_factor in accessor_conjugate
  }

  offset_policy::pointer offset(pointer p, ptrdiff_t i) const noexcept {
    return a.offset(p,i);
  }

  element_type* decay(pointer p) const noexcept {
    return a.decay(p);
  }
private:
  Accessor acc;
};

```

mhoemmen:
@p-groarke Thanks for the catch!  I'd like to merge this PR first and then fix it, so that we have a commit with the state as submitted.  I'll be sure to fix this in the next revision, though!  (Also, the two papers didn't get included in the initial web list -- I'm talking to Hal about that.)

mhoemmen:
FYI the papers got submitted correctly.

mhoemmen:
@crtrott @hcedwar @dhollman @dsunder

PR #120, opened 2020-01-03
P1673R2: Check compile-time extents via Mandates clause

mhoemmen:
@crtrott @dalg24

### DIFF
--- a/D1673/blas_interface.md
+++ b/D1673/blas_interface.md
@@ -2015,6 +2015,10 @@ this.
     domain of `y`, the expressions `x(i) = c*x(i) + s*y(j)` and
     `y(j) = c*y(j) - conj(s)*x(i)` are well formed.
 
+* *Mandates:* If neither `x.static_extent(0)` nor `y.static_extent(0)`
+  equals `dynamic_extent`, then `x.static_extent(0)` equals
+  `y.static_extent(0)`.
+
 * *Effects:* Applies the plane (Givens) rotation specified by `c` and
   `s` to the input vectors `x` and `y`, as if the rotation were a 2 x
   2 matrix and the input vectors were successive rows of a matrix with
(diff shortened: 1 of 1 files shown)


D1673/blas_interface.md
```diff
@@ -60,6 +60,9 @@
 
   * Removed over- and underflow requirement for `vector_norm2`.
 
+  * If any matrix or vector dimension compatibility checks can be done
```
dalg24: Maybe phrase it "Mandate that"

D1673/blas_interface.md
```diff
@@ -2003,8 +2006,6 @@ this.
 
   * `Real` is `float`, `double`, or `long double`.
 
-  * `x.rank()` and `y.rank()` are both one.
-
```
dalg24: Did you mean to remove the constraint on their ranks?

D1673/blas_interface.md
```diff
@@ -2178,13 +2207,19 @@ real element types), `xDOTC`, and `xDOTU` (for complex element types).
 * *Requires:*
 
   * `T` shall be *Cpp17MoveConstructible*.
+
   * `init + v1(0)*v2(0)` shall be convertible to `T`.
-  * `v1` and `v2` have the same domain.
```
dalg24: Did you mean to replace all occurrences of "have the same domain"?
I am asking because I noticed there is at least one in `givens_rotation_apply()`

D1673/blas_interface.md
mhoemmen: Yes -- that constraint was already imposed above, as one of the requirements on template parameters with "vector" in their name.

D1673/blas_interface.md
mhoemmen: @dalg24 I'm still thinking about that ;-) .  The reason I changed the other occurrences was to have the Requires and Mandates clauses match, so that it would be easier to read the Mandates clauses.  I could also replace the other instances of "have the same domain."

mhoemmen: @dalg24 Christian thinks we should change the "have the same domain" things first.  I'll do that.

D1673/blas_interface.md
mhoemmen: Changed -- thanks!

D1673/blas_interface.md
mhoemmen: Done!

PR #123, opened 2020-01-06
P1673R2: Changes based on 2020 Jan 03 review

mhoemmen:
@crtrott @dalg24

### DIFF
--- a/D1673/blas_interface.md
+++ b/D1673/blas_interface.md
@@ -1762,17 +1762,51 @@ appropriate run-time BLAS parameters.
 
 #### `layout_transpose`
 
-This layout wraps an existing layout, and swaps its rightmost two
-indices.
+`layout_transpose` is a `basic_mdspan` layout mapping policy that
+swaps the rightmost two indices, extents, and strides (if applicable)
+of an existing unique layout mapping policy.
 
 ```c++
 template<class InputExtents>
 using transpose_extents_t = /* see below */; // exposition only
+```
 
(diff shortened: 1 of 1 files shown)


D1673/blas_interface.md
```diff
@@ -1691,160 +1808,337 @@ public:
...
 
+* *Requires:*
+
+  * `Layout` shall meet the `basic_mdspan` layout mapping policy
+    requirements (see *[mdspan.layout.reqs]* in P0009).
+
 * *Constraints:*
 
-  * `Layout` is a unique layout.
+  * `Layout::is_always_unique()` is `true`.
```
crtrott: Why do we need this? I would think we can just inherit this from the nested layout?

mhoemmen: @crtrott Ah, good point; thanks!

crtrott: @mhoemmen you want to change that first? Or shall I just merge this in and then we can change that later?

mhoemmen: @crtrott I'll just change it later -- I'll need to make a bunch of changes for `layout_blas_*` anyway.  Thanks!

mhoemmen: @crtrott Oh, wait, I get it.  `Layout` _is_ the nested layout.  This is a constraint on the nested layout.  I still need to fix this, though, because the nested layout does not define this -- the nested layout's mapping does.

PR #124, opened 2020-01-07
P1673R2: 2020 Jan 06 wording changes

mhoemmen:
@crtrott @dalg24 @dhollman et alia: Most important changes:

  * Finished wording for `layout_blas_general` and `layout_blas_packed`
  * Moved or removed most non-wording introductory material out of the wording section
  * Made explicit where the wording section starts, by putting wording in a new section called "Wording"
  * (new as of 2020 Jan 07 AM) finished adding a synopsis and section headers

### DIFF
--- a/D1673/blas_interface.md
+++ b/D1673/blas_interface.md
@@ -1345,6 +1345,16 @@ alpha*x + beta*y`:
 linalg_add(scaled_view(alpha, x), scaled_view(beta, y), w);
 ```
 
+*[Note:*
+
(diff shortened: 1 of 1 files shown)


mhoemmen:
I just added a new commit that finishes the synopsis and section labels.  I might end up changing the section labels to make the naming scheme more consistent, but for now at least all the wording is there.

mhoemmen:
@crtrott @dhollman @dalg24 et alia, I'm done with the current round of changes.  I will write up a poll for the next round.  Thanks all!

PR #126, opened 2020-01-10
P1673R2: Fix *_view return type, etc. (10 Jan 2020)

mhoemmen:
@crtrott @dhollman @dalg24 @dsunder and friends :-)

The most important change is that the return type of `*_view` is now "see-below".  Implementations have freedom to optimize multiple applications of `layout_transpose`, `accessor_scaled`, `accessor_conjugate`, etc., but are not required to.  I added some Notes to explain this.


mhoemmen:
Use #127

PR #139, opened 2021-05-18
P0009R12 mdspan

nliber:
R12 changes:

- Fixed definition of `static_extent`
- Added converting constructor for `default_accessor` (when the pointers to
  elements are convertible) for things like `default_accessor<double>` to `default_accessor<const double>`
- Changed [mdspan.accessor.basic] to [mdspan.accessor.default], to correspond with the name change
  in P0009r11
- Minor formatting corrections


keryell:
2 versions of P0009 per day? Lucky time! ;-)

PR #140, opened 2021-06-15
Remove all references to ptrdiff_t in stdBLAS

amklinv-nnl:
ptrdiff_t was the type used by mdspan to store sizes/extents. Now that mdspan has moved to size_t instead, we use the mdspan defined size_type. Note that some functions claim to return a -1, which they cannot do anymore (kokkos/stdBLAS#73). Closes kokkos/stdBLAS#74.


crtrott:
Was also part of #142

PR #148, opened 2021-08-11
P1673R4

mhoemmen:


D1673/P1673.bs
```diff
@@ -1401,9 +1401,10 @@ The types of `x(i)` and `y(i)` might be more complicated;
 e.g., `x(i)` might be a polynomial with `double` coefficients,
 and `y(i)` a polynomial with `float` coefficients.
 If those polynomials use expression templates,
-then the expression `x(i) + x(i)` might have a completely different type
-than `decltype(x(i))` (possibly with references removed),
-and might also have a completely different type than `alpha * x(i) + y(i)`.
+then slightly different sum expressions involving `x(i)` and/or `y(i)`
+(e.g., `alpha * x(i) + y(i)`, `x(i) + y(i)`, or `y(i) + x(i)`)
+might all have different types,
+all of which differ from value type of `x` or `y`.
```
opensdh, line 1407: It's probably worth emphasizing that all these types still have to be convertible (or implicitly convertible, or assignable?) to the output type(s).

D1673/P1673.bs
```diff
@@ -1681,14 +1679,21 @@ that create significant temporary storage
...
 yet both parallelization and customary linear algebra optimizations
 rely on reordering sums.
-See the above "adverbs, not adjectives" section for a more detailed explanation.
+See the above "Value type constraints do not suffice to describe algorithm behavior" section
+for a more detailed explanation.
+
+Regarding Assumption (3),
+we do not actually say that
+value-initialization produces a two-sided additive identity.
+What matters is what the algorithm's implementation may assume,
```
opensdh, line 1695: Using "assume" here isn't as much of a problem as it would be in the wording, but "do" would work just as well.

D1673/P1673.bs
```diff
@@ -5088,61 +5064,77 @@ satisfy the following requirements.
...
+    resulting from a valid output `mdspan` access.
+
+5. If the algorithm's mathematical expression
+    uses addition,
+    but not subtraction, multiplication, or division,
+    then the algorithm evaluates the mathematical expression
+    using a sequence of evaluations of
+    `+`, `+=`, and `=` operators
+    that would produce the correct result
+    when operating on elements of a commutative semigroup.
```
opensdh, line 5089: I think "exact arithmetic" _would_ work here: it doesn't require that implementations _use_ exact arithmetic (or test with it), because we know from algebra what transformations would preserve the value of the "textbook expression".  However, using the algebraic terminology might be superior in that it specifies the permissible operations other than implicitly via the allowed operators.

D1673/P1673.bs
```diff
@@ -5088,61 +5064,77 @@ satisfy the following requirements.
...
+    `*`, `*=`, `+`, `+=`,
+    unary `-`, binary `-`, `-=`,
+    and `=` operators
+    that would produce the correct result
+    when operating on elements of a ring
+    with noncommutative multiplication.
+
+8. If the algorithm's mathematical expression
+    uses division and possibly also
+    addition, subtraction, and multiplication,
```
opensdh, line 5117: You could reverse these to simplify the descriptions:

> 1. If the algorithm's mathematical expression _E_ uses division, then the algorithm evaluates _E_ using…
> 1. Otherwise, if _E_ uses subtraction, then the algorithm evaluates _E_ using…
> 1. Otherwise, if _E_ uses multiplication, …

D1673/P1673.bs
```diff
@@ -5088,61 +5064,77 @@ satisfy the following requirements.
...
+    and `=` operators
+    that would produce the correct result
+    when operating on elements of a field
+    with noncommutative multiplication.
+
+9. The algorithm may reorder addends in its mathematical expression arbitrarily.
     *[Note:*
-    The algorithm may *not* reorder products arbitrarily.
+    Factors in each product are not reordered.
     This is because the algorithms do not assume that multiplication is commutative.
```
opensdh, line 5131: I'd omit this or reduce it to a suffix of the previous sentence:

> ; multiplication need not be commutative.

D1673/P1673.bs
mhoemmen: @opensdh Thanks for the feedback!  That's a good point.

D1673/P1673.bs
mhoemmen: > I think "exact arithmetic" would work here....

In this particular case of only `+`, `+=`, and `=`, I think it would.  I've been trying to avoid that phrase for anything involving `*`, because it might incorrectly suggest the commutative multiplication of real numbers.

D1673/P1673.bs
mhoemmen: Ah, that's an excellent suggestion; thank you!  I might also add that we interpret "a / b" as "a times the inverse of b", to avoid confusion in the noncommutative case.

D1673/P1673.bs
mhoemmen: Fixed; thanks!

D1673/P1673.bs
mhoemmen: I think we need both `assignable_from` and `convertible_to` here.  I don't want implementations to have to create a new `output_value_type` instance just to be able to assign it to the output `mdspan`.

D1673/P1673.bs
mhoemmen: Fixed; thanks!

D1673/P1673.bs
mhoemmen: Fixed; thanks!

D1673/P1673.bs
mhoemmen: I might actually want to do the opposite here: "If the algorithm's mathematical expression does _not_ use subtraction, then...."  I'm as much trying to exclude non-textbook algorithm choices (like Strassen's algorithm, for types that aren't all floating-point types), as I'm trying to show all permitted operations.  What do you think?

opensdh: My suggestion preserves meaning; yours does only if every expression uses a set of operations that is a prefix of `+*-/`.

mhoemmen: Fixed in the latest draft; thanks for the suggestion!

mhoemmen:
@crtrott @dalg24 @amklinv-nnl The P1673R4.html file is the R4 version that I submitted for this month's mailing.

mhoemmen:
Thanks Christian!

PR #150, opened 2021-09-25
Add back in design discussion for P0009 as requested by LEWG

crtrott:


dalg24 (approved):

P0009/P0009r12.md
```diff
@@ -695,13 +704,123 @@ array functionality of this proposal. Two critical functionality
...
+One important reason for this is that subspans enable code reuse.
+For example the inner loop in the above described matrix-vector product actually 
+represents a *dot product* - an inner product of two vectors.
+If one already has a function for such an inner product, then using it by obtaining 
+each row of the matrix as a subspan allows its reuse.
+
+<b> Reference Implementation </b>
+
+A reference implementation under BSD license is available at:
+[mdspan](https://github.com/kokkos/mspan).
```
nliber, line 821: The link should be https://github.com/kokkos/mdspan

nliber (changes requested):

mhoemmen:
@dalg24 @nliber -- @crtrott and I were planning for me to make another pass on this for grammar and style.  I can either do that in a separate PR based on this one, or we can fix the issues you pointed out first and merge.  It's up to y'all : - )  We do need to get this done soon to make the mid-October deadline.

dalg24:
Your choice.  I approved regardless.

nliber:
Fix the link and I approve this as well.

Assuming P2128 gets approved for the WD on Monday, we still have:

> If [P2128 Multidimensional subscript operator](https://wg21.link/P2128/) is also adopted into C++23, we will likely
> propose that `basic_mdspan::operator()` be replaced with `basic_mdspan::operator[]` for C++23.

I would rather we support both operator() and operator[] so that we can use mdspan with pre-C++23 code (Kokkos and SYCL).

nliber:
> @nliber wrote:
> 
> > I would rather we support both operator() and operator[] so that we can use mdspan with pre-C++23 code (Kokkos and SYCL).
> 
> Implementations of mdspan could certainly provide this. Do you think it's worth standardizing, or would we expect vendors just to provide the nonstandard extension (since using mdspan with pre-C++23 code would be nonstandard anyway)?

I think it is worth standardizing.  Otherwise, users have to change code when they get to C++23 (I'm not sure vendors can still provide it as an extension and be conformant to C++23) as they will have no evolutionary path.  LEWG may, of course, think differently.

mhoemmen:
@nliber Let's talk more offline -- thanks!

crtrott:
@mhoemmen are you allowed to push into this? If not feel free to open a new PR based on this.

mhoemmen:
@crtrott I just tried a push -- it doesn't work, since I don't have push permission to your fork.

mhoemmen:
Per discussion with @crtrott , this is superseded by PR https://github.com/ORNL/cpp-proposals-pub/pull/154 .  Please close; thanks!

PR #154, opened 2021-10-11
P0009: Improve description & justification

mhoemmen:
@crtrott @dalg24 @nliber please review; thanks! : - )

I'm not entirely done, but this should be fine to merge as-is.  This supersedes PR https://github.com/ORNL/cpp-proposals-pub/pull/150 .

### DIFF
--- a/P0009/P0009r12.md
+++ b/P0009/P0009r12.md
@@ -719,7 +719,7 @@ It also proposes adding the `submdspan` function that "slices"
 (returns an `mdspan` that views a subset of) an existing `mdspan`.
 
 The `mdspan` class template can represent arbitrary mixes
-of compile-time or run-time dimensions.
+of compile-time or run-time extents.
 Its element type can be any complete object type
 that is neither an abstract class type nor an array type.
 It has two customization opportunities for users:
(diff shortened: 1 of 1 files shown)


P0009/P0009r12.md
```diff
@@ -708,89 +709,180 @@ properties will work.
...
+A pointer-of-pointers also loses information
+about any dimensions known at compile time.
+Users cannot arbitrarily mix compile-time and run-time dimensions.
+
+Approach (3) can mix `vector` and `array` to represent dimensions
+known at run time resp. compile time.
+However, a `vector` of `vector`s (of `vector`s) is a container
+with many separate allocations.
+It cannot represent a view of contiguous memory.
+
```
crtrott, line 810: Add: both (3) and (4) also have significant performance issues, since for each access to a value multiple memory locations must be read, i.e. pointer chasing occurs.

P0009/P0009r12.md
```diff
@@ -708,89 +709,180 @@ properties will work.
...
+These can significantly improve the generated code.
+Not storing dimensions at run time
+may help conserve registers and stack space.
+
+In many fields, some dimensions are naturally known at compile time.
+For many physics and engineering algorithms,
+some dimensions are dictated by fundamental properties of the physical world
+or the discretization scheme.
+For example, the position of a particle in space
+requires a three-dimensional array, since physical space has three dimensions.
```
crtrott, line 840: don't say three-dimensional array here. I very explicit used "array of extent three" so it doesn't get confused with the other use of dimension. In particular her you say we represent a "three-dimensional array" with "other dimensions" as a "two-dimensional array"

crtrott (changes requested):

P0009/P0009r12.md
mhoemmen: Fixed; thanks!

PR #162, opened 2021-10-20
P0009: add missing instructions to submdspan for layout type

crtrott:
also how to initialize accessor and pointer.


P0009/P0009.md
```diff
@@ -2612,7 +2612,7 @@ namespace std {
   // [mdspan.submdspan], submdspan creation
   template<class ElementType, class Extents, class LayoutPolicy,
            class AccessorPolicy, class... SliceSpecifiers>
-      constexpr mdspan<@_see below_@>
+      constexpr mdspan<ElementType, @_see below_@, @_see below_@, typename AccessorPolicy::offset_policy>
```
mhoemmen, line 2615: This is much nicer -- I like that we provide all the template arguments here.

PR #166, opened 2021-11-03
Make constructor conditionally explicit

crtrott:
- makes mdspan, mapping and accessor constructors conditionally explicit
- makes extents conversion constructor explicit for dynamic -> static
- fixes the pointer derived to base class assignment for
  default_accessor

### DIFF
--- a/P0009/P0009.md
+++ b/P0009/P0009.md
@@ -1304,7 +1304,7 @@ public:
   constexpr extents& operator=(extents&&) noexcept = default;
 
   template<size_t... OtherExtents>
-    explicit((((Extents!=dynamic_extent) && (OtherExtents==dynamic_extent)) || ... ))
+    explicit(@_see below_@)
     constexpr extents(const extents<OtherExtents...>&) noexcept;
   template<class... SizeTypes>
     explicit constexpr extents(SizeTypes...) noexcept;
@@ -1366,7 +1366,7 @@ equivalent to `dynamic_extent`.  Otherwise, `rank_dynamic()`.
 
 ```c++
 template<size_t... OtherExtents>
-  explicit((((Extents!=dynamic_extent) && (OtherExtents==dynamic_extent)) || ... ))
+  explicit(@_see below_@)
   constexpr extents(const extents<OtherExtents...>& other) noexcept;
 ```
 
@@ -1385,6 +1385,8 @@ template<size_t... OtherExtents>
 * [3]{.pnum} *Effects:* For each `r` where `static_extent(r) == dynamic_extent` is `true`,
              assigns `other.extent(r)` to `dynamic_extent[dynamic_index(r)]`.
 
+* [4]{.pnum} *Remarks:* The expression inside `explicit` is equivalent to:
+             `(((Extents!=dynamic_extent) && (OtherExtents==dynamic_extent)) || ... )`
 <br/>
 
 ```c++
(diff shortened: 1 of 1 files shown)


crtrott:
gotdbolt mockup to play with the mechanics, and check out whether these are the right type of changes
https://godbolt.org/z/68vK9zM7T

Still need to work on the implementation itself.

P0009/P0009.md
```diff
@@ -1299,9 +1299,10 @@ public:
   constexpr extents& operator=(extents&&) noexcept = default;
 
   template<size_t... OtherExtents>
+    explicit((((Extents!=dynamic_extent) && (OtherExtents==dynamic_extent)) || ... ))
```
dalg24, line 1302: extra pair of parenthesis?

crtrott: (try the godbolt above)

dalg24 (commented):
You might to update the changelog

crtrott:
Added changelog

mhoemmen:
We plan to address default constructibility of `mdspan` (i.e., to make it require that the relevant members are default constructible) in a subsequent PR, in order to avoid confusion.

P0009/P0009.md
```diff
@@ -2131,10 +2131,7 @@ a contiguous set of objects are accessed.
   * [2.2]{.pnum} a handle to a contiguous set of elements of type `element_type`,
     accessible through the policy's `access` method;
 
-  * [2.3]{.pnum} conversion of a handle to a contiguous set of elements,
-      to a pointer _[conv.array]_; and
-
-  * [2.4]{.pnum} getting a handle to the contiguous subset of elements
+  * [2.3]{.pnum} getting a handle to the contiguous subset of elements
```
mhoemmen, line 2134: Should we take the opportunity to reword all these bullets?  It's a bit inconsistent how the other bullet points use "a handle to."  2.2 also refers to a "contiguous set of elements," but we don't promise contiguity.  Here's my attempt:

```
* [2.1]{.pnum} the type of a handle to a single element of type `element_type`;

* [2.2]{.pnum} a method to access the `i`-th element in a range of elements; and

* [2.3]{.pnum} a method to access the subrange beginning at an integer offset value of a given range of elements.
```
In any case, we do need "; and" at the end of 2.2 now, given that it's the penultimate point.

P0009/P0009.md
```diff
@@ -2322,7 +2319,7 @@ public:
   using reference = typename accessor_type::reference;
 
   // [mdspan.basic.cons], mdspan constructors, assignment, and destructor
-  constexpr mdspan() = default;
+  constexpr mdspan() requires(rank_dynamic != 0) = default;
```
mhoemmen, line 2322: This differs from `span` in that it permits default construction if any extent is (statically) zero.  Is that intentional?  I'm guessing so (I would want an `mdspan` with any static extent of zero to be default constructible), but just checking.

P0009/P0009.md
```diff
@@ -2559,19 +2556,19 @@ template<class OtherElementType, class OtherExtents, class OtherLayoutPolicy, cl
...
-    + [13.4]{.pnum} `OtherExtents::rank() == rank()` is `true`; and
+    + [13.3]{.pnum} `OtherExtents::rank() == rank()` is `true`; and
 
-    + [13.5]{.pnum} For all `r` in the range `[0, rank())`, 
+    + [13.4]{.pnum} For all `r` in the range `[0, rank())`, 
        if `other.static_extent(r) != dynamic_extent && static_extent(r) != dynamic_extent` is `true`,
        then `other.static_extent(r) == static_extent(r)` is `true`.
 
-* [14]{.pnum} *Preconditions:* For all `r` in the range `[0, rank())`,
+* [14]{.pnum} *Mandates:* `is_assignable_v<pointer, OtherAccessor::pointer>` is `true`.
```
mhoemmen, line 2565: I do prefer this change -- it's nicer to have the check here, rather than in the depths of the accessor.  Thanks! : - )

P0009/P0009.md
crtrott: ah hm ok I can change that.

P0009/P0009.md
crtrott: yeah we should look at some of the other words to see if there is more stuff which needs to rewording. I agree that the previous wording was clunky. Do you want to issue a PR with your suggestion? Though 2.3 is not about access to a subrange, but more about creating a subrange.

mhoemmen: @crtrott I'll create a PR with these changes; thanks!

mhoemmen: PR with these changes and a few more: https://github.com/ORNL/cpp-proposals-pub/pull/171

PR #167, opened 2021-11-08
P0009: Add more design justification

mhoemmen:
@crtrott Add more design justification to P0009.

### DIFF
--- a/P0009/P0009.md
+++ b/P0009/P0009.md
@@ -861,7 +861,9 @@ and is meant for array-based computation.
 Accessing a single element requires accessing
 the memory of three `valarray`s.
 
-## Mixing compile-time and run-time extents
+## Extents
+
+### Mixing compile-time and run-time extents
 
 The fundamental reason to allow expressing extents at compile time is performance.
 Knowing an extent at compile time enables many compiler optimizations,
@@ -889,6 +891,52 @@ might only be known at run time.
 This would make a rank-3 array with two compile-time extents
 a natural data structure for the matrices.
 
+Different extents will result in different `mdspan` instantiations,
+even if the element type, rank, layout, and accessor are all the same.
+Some users may be concerned that this will cause a proliferation of types.
+Users with this concern can restrict themselves to only dynamic extents
+and the default layout and accessor.
+They can also "type erase" all three layouts we propose here
+to a single layout type, `layout_stride`.
+
+Our view is that forcing too much type erasure
+(e.g., only permitting dynamic extents,
+or making layout or accessor run-time polymorphic) would hinder performance.
+C++ has generally been moving towards providing more information at compile time.
+Our experience is that compilers optimize array access better
+when given more information about strides at compile time.
+
+We made `mdspan`'s element type, extents, layout, and accessor orthogonal
+in order to simplify customization.
+Even if changing `mdspan`'s design to mix these policies together
+would reduce the number of instantiations,
+we would not do it, because it would make customization harder.
+
+### Conversions between run-time and compile-time extents
+
+We permit converting between `mdspan` types
+that have the same rank, layout, and accessor,
+as long as the extents match at run time.
+This includes both conversion of static extents to dynamic extents,
+and conversion of dynamic extents to static extents.
+Both of these cases are important for helping users made trade-offs
+between compile times and sizes, and run-time performance.
+Static-to-dynamic conversion is a kind of type erasure,
+for when an implementation does not specialize for compile-time extents.
+Dynamic-to-static conversion is a kind of specialization,
+for when an implementation is willing to pay the compile-time cost of specializing.
+One use case for dynamic-to-static conversion is in applications
+that solve partial differential equations through a variety of discretization methods.
+Many methods result in arrays with known fixed extents,
+but making all the extents compile-time constants might cost too much compilation time.
+A rational approach is to make just the most commonly occurring extents compile-time constants.
+
+Converting extents from dynamic to static can cause precondition violations.
+Since this is an important use case, we permit such `mdspan` conversions,
+but we require them to be explicit.
+Static-to-dynamic conversion cannot cause precondition violations,
+so we permit implicit `mdspan` conversions in that case.
+
 ## Why custom memory layouts?
 
 Our `mdspan` class template permits custom layouts.
@@ -914,8 +962,6 @@ Control of the layout can also be used to write code
 that performs well on different computer architectures
 when only changing a template argument.
 Consider the following implementation of a parallel dense matrix-vector product.
-
-
 
 ```c++
 using layout = /* see-below */;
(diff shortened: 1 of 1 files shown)


P0009/P0009.md
```diff
@@ -889,6 +891,52 @@ might only be known at run time.
...
+or making layout or accessor run-time polymorphic) would hinder performance.
+C++ has generally been moving towards providing more information at compile time.
+Our experience is that compilers optimize array access better
+when given more information about strides at compile time.
+
+We made `mdspan`'s element type, extents, layout, and accessor orthogonal
+in order to simplify customization.
+Even if changing `mdspan`'s design to mix these policies together
+would reduce the number of instantiations,
+we would not do it, because it would make customization harder.
```
crtrott, line 913: this doesn't make much sense. How does mixing reduce instantiations if you don't reduce the amount of information encoded?

P0009/P0009.md
```diff
@@ -889,6 +891,52 @@ might only be known at run time.
...
+
+### Conversions between run-time and compile-time extents
+
+We permit converting between `mdspan` types
+that have the same rank, layout, and accessor,
```
crtrott, line 918: Not true: we allow conversion as long as mappings, accessors, and pointers are convertible. And the matching of extents is not true either, since the left hand side might be uninitialized and hence its dynamic extents are 0.

P0009/P0009.md
```diff
@@ -889,6 +891,52 @@ might only be known at run time.
...
+Our experience is that compilers optimize array access better
+when given more information about strides at compile time.
+
+We made `mdspan`'s element type, extents, layout, and accessor orthogonal
+in order to simplify customization.
+Even if changing `mdspan`'s design to mix these policies together
+would reduce the number of instantiations,
+we would not do it, because it would make customization harder.
+
+### Conversions between run-time and compile-time extents
```
crtrott, line 915: Not sure we need this at all: this just matches std::span now.

P0009/P0009.md
```diff
@@ -889,6 +891,52 @@ might only be known at run time.
...
+as long as the extents match at run time.
+This includes both conversion of static extents to dynamic extents,
+and conversion of dynamic extents to static extents.
+Both of these cases are important for helping users made trade-offs
+between compile times and sizes, and run-time performance.
+Static-to-dynamic conversion is a kind of type erasure,
+for when an implementation does not specialize for compile-time extents.
+Dynamic-to-static conversion is a kind of specialization,
+for when an implementation is willing to pay the compile-time cost of specializing.
+One use case for dynamic-to-static conversion is in applications
```
crtrott, line 928: not sure if this example is good. Isn't like graphics even better and more widely known?

P0009/P0009.md
```diff
@@ -889,6 +891,52 @@ might only be known at run time.
...
+that solve partial differential equations through a variety of discretization methods.
+Many methods result in arrays with known fixed extents,
+but making all the extents compile-time constants might cost too much compilation time.
+A rational approach is to make just the most commonly occurring extents compile-time constants.
+
+Converting extents from dynamic to static can cause precondition violations.
+Since this is an important use case, we permit such `mdspan` conversions,
+but we require them to be explicit.
+Static-to-dynamic conversion cannot cause precondition violations,
+so we permit implicit `mdspan` conversions in that case.
```
crtrott, line 938: Just saying we follow std::span conventions here should be good enough.

P0009/P0009.md
```diff
@@ -1011,6 +1059,70 @@ for supporting heterogeneous memory.
...
+
+Our design orthogonalizes storage type or allocation handle (the pointer)
+from access properties (the accessor).
+An `mdspan` stores a `pointer`,
+but uses its accessor for element access and for making offset views.
+Just as `atomic_ref` and ordinary pointer dereferencing
+can access the same memory in two different ways,
+two different `mdspan` objects might access the same memory in two different ways.
+
+This explains why the accessor policy uses `a.access(p, i)`
```
crtrott, line 1072: why mdspan uses a.access(p,i). Also this duplicates information from above. better to merge.

P0009/P0009.md
```diff
@@ -1011,6 +1059,70 @@ for supporting heterogeneous memory.
...
+`mdspan` stores its pointer separately from the accessor.
+This has the following advantages:
+
+1. one can construct two different `mdspan`s with the same pointer,
+    even if they have different accessor types
+    (with the same `pointer` type);
+
+2. the accessor need not be stateful;
+
+3. `mdspan` is default constructible, even if the accessor is not; and
```
crtrott, line 1091: ?? It is not since the accessor member needs to be default constructible for mdspan to be.

P0009/P0009.md
```diff
@@ -1011,6 +1059,70 @@ for supporting heterogeneous memory.
...
+    instances of two different accessor types that want to access the same memory.
+
+Regarding (1): We want it to be possible for two different `mdspan`
+to access the same memory in different ways.
+
+Regarding (3): `mdspan` is default constructible
+only if its layout and accessor types are default constructible.
+This permits stateful accessors that might not make sense to default construct.
+However, the `pointer` type needs to be default constructible,
+else `mdspan` would not be trivially constructible.
```
crtrott, line 1103: it is anyway not, as we figured out, since value initalization makes it not trivially constructible.

P0009/P0009.md
```diff
@@ -1011,6 +1059,70 @@ for supporting heterogeneous memory.
...
+Just as `T*` and `span<T>` are default constructible,
+`mdspan` needs to be default constructible.
```
crtrott, line 1105: but we just said that its not if the accessor type is not.

crtrott (changes requested):
I am generally not sure we need all of this. Definitely some of it is not correct, and others need to be merged with the previous subsection (in accessor)

P0009/P0009.md
mhoemmen: Right, exactly : - )  I'm not saying it makes sense, rather that even if it were somehow possible, it wouldn't be what we want.

P0009/P0009.md
```diff
@@ -889,6 +891,52 @@ might only be known at run time.
...
+for when an implementation does not specialize for compile-time extents.
+Dynamic-to-static conversion is a kind of specialization,
+for when an implementation is willing to pay the compile-time cost of specializing.
+One use case for dynamic-to-static conversion is in applications
+that solve partial differential equations through a variety of discretization methods.
+Many methods result in arrays with known fixed extents,
+but making all the extents compile-time constants might cost too much compilation time.
+A rational approach is to make just the most commonly occurring extents compile-time constants.
+
+Converting extents from dynamic to static can cause precondition violations.
```
crtrott, line 934: As in std::span converting from dynamic to static can cause precondition violations.

mhoemmen:
Closing; this needs more work, and isn't urgent.  Thanks all!

PR #171, opened 2021-11-12
P0009: Fix accessor policy; stop implying "contiguous"-ness

mhoemmen:
@crtrott @dalg24 @nliber 

1. Fix accessor policy wording, in particular the bulleted list
   that wasn't lining up with the table.

2. Remove implications that the set of elements that an accessor
   accesses is contiguous in memory.  (It need not be.)

### DIFF
--- a/P0009/P0009.md
+++ b/P0009/P0009.md
@@ -779,7 +779,7 @@ and whose upper bound is the product of the <math>R</math> extents.
 
 More formally, a multidimensional array of rank <math>R</math>
 maps from its *domain*, a multidimensional index space of rank <math>R</math>,
-to its *codomain*, elements of a contiguous span of objects.
+to its *codomain*, a set of objects accessible from a contiguous range of integer indices.
 A *multidimensional index space* of rank <math>R</math>
 is the Cartesian product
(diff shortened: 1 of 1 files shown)


trws:
To be sure I'm reading this right, the elements in memory need not be contiguous, but the range of indices needs to be a logically contiguous rectangular space?

mhoemmen:
@trws wrote:
> To be sure I'm reading this right, the elements in memory need not be contiguous, but the range of indices needs to be a logically contiguous rectangular space?

That's right!  The domain of an `mdspan` is a Cartesian product of `[0, extent(k))` for `k` in `[0, rank())`.  The layout maps a multidimensional index to a 1-D index.  The accessor maps from the `mdspan`'s pointer and 1-D index, to a reference to an element.  We do not require the pointer type to be `element_type*`, nor do we require that the elements be stored contiguously in memory.  (They need not be in memory at all.  Consider PGAS accessors, for example.)

mhoemmen:
This is not a new change -- we had removed the assumption of an underlying `span` by R11.

trws:
Right, that makes sense. Thanks Mark.
________________________________

mhoemmen:
Gladly Tom! : - D

crtrott (approved):

PR #174, opened 2021-11-12
Allow rank args

crtrott:


P0009/P0009.md
```diff
@@ -54,6 +54,7 @@ parameters are only explicitly convertible
     changing rank or assigning different sized static extents
 - add feature test macro
 - fix layout_stride constructor to be flexible with integral types of strides array
```
mhoemmen, line 56:
```suggestion
- fix `layout_stride` constructor to be flexible with integral types of strides array
```

P0009/P0009.md
```diff
@@ -54,6 +54,7 @@ parameters are only explicitly convertible
...
+- make extents and mdspan constructors accept either rank_dynamic or rank integer args (or array of that size)
```
mhoemmen, line 57:
```suggestion
- make `extents` and `mdspan` constructors accept either `rank_dynamic` or `rank` integer arguments (or an `array` of that size)
```

P0009/P0009.md
```diff
@@ -1355,28 +1357,37 @@ of rank equal to `sizeof...(Extents)`.
 
 [2]{.pnum} `extents<Extents...>` is a trivially copyable type.
 
-[3]{.pnum} $E_r$ is a _dynamic extent_ if it is equal to `dynamic_extent`, otherwise $E_r$ is a _static extent_.
+[3]{.pnum} Let $E_r$ be the $r^{th}$ element of $Extents$.
```
mhoemmen, line 1360: This is helpful; nice catch! : - )

P0009/P0009.md
```diff
@@ -1355,28 +1357,37 @@ of rank equal to `sizeof...(Extents)`.
...
+
+[4]{.pnum} $E_r$ is a _dynamic extent_ if it is equal to `dynamic_extent`, otherwise $E_r$ is a _static extent_.
 For each $E_r$ equivalent to `dynamic_extent`,
```
mhoemmen, line 1363:
```suggestion
For each $E_r$ that is a dynamic extent,
```
1. It's not clear why we need to say "equivalent" here.
2. It would be nicer to use the definition on line 1362 immediately after using it.

P0009/P0009.md
```diff
@@ -1355,28 +1357,37 @@ of rank equal to `sizeof...(Extents)`.
...
+[3]{.pnum} Let $E_r$ be the $r^{th}$ element of $Extents$.
+
+[4]{.pnum} $E_r$ is a _dynamic extent_ if it is equal to `dynamic_extent`, otherwise $E_r$ is a _static extent_.
 For each $E_r$ equivalent to `dynamic_extent`,  
 the upper bound of the interval is stored in the exposition-only array `dynamic_extents_` at
 `dynamic_extents_[dynamic_index(`_r_`)]`.
 
-[4]{.pnum} If $E_r$ is a dynamic extent, let $D_r$ be the value of `dynamic_extents_[dynamic_index(`_r_`)]`.
+[5]{.pnum} If $E_r$ is a dynamic extent, let $D_r$ be the value of `dynamic_extents_[dynamic_index(`_r_`)]`.
 The $r^{th}$ interval of an `extents` is as follows:
```
mhoemmen, line 1368:
```suggestion
The $r^{th}$ interval of an `extents` object is as follows: 
```

P0009/P0009.md
```diff
@@ -1355,28 +1357,37 @@ of rank equal to `sizeof...(Extents)`.
...
 
 ---
 
 ```c++
 constexpr size_t dynamic_index(size_t i) noexcept; // @_exposition only_@
 ```
 
-* [5]{.pnum} *Returns*: If `i <= sizeof...(Extents)` is `true`, the number of arguments 
-before the `i`_th_ template parameter in the template parameter pack `Extents` 
+* [7]{.pnum} *Returns*: If `i <= sizeof...(Extents)` is `true`, the number of arguments
```
mhoemmen, line 1380:
```suggestion
* [7]{.pnum} *Returns*: If `i <= sizeof...(Extents)` is `true`, then the number of arguments
```

P0009/P0009.md
```diff
@@ -1407,26 +1418,39 @@ template<size_t... OtherExtents>
...
 ```
 
-* [4]{.pnum} *Constraints:*
+* [5]{.pnum} *Constraints:*
     
-    * [4.1]{.pnum} `(is_convertible_v<SizeTypes, size_type> && ...)` is `true`, and
+    * [5.1]{.pnum} `(is_convertible_v<SizeTypes, size_type> && ...)` is `true`, and
     
-    * [4.2]{.pnum} `sizeof...(SizeTypes) == rank_dynamic()` is `true`.
+    * [5.2]{.pnum} `sizeof...(SizeTypes)` is `dynamic_rank()` or `rank()`.
```
mhoemmen, line 1428:
```suggestion
    * [5.2]{.pnum} `sizeof...(SizeTypes)` equals either `dynamic_rank()` or `rank()`.
```

P0009/P0009.md
```diff
@@ -1407,26 +1418,39 @@ template<size_t... OtherExtents>
...
 
-* [5]{.pnum} *Effects:*  Initializes `dynamic_extents_` with `dynamic...`.
+* [5]{.pnum} *Preconditions:* If `sizeof...(SizeTypes) == rank()` is `true`,
+             $r$ is smaller than `sizeof()...SizeTypes)`, and
```
mhoemmen, line 1431:
```suggestion
             $r$ is smaller than `sizeof...(SizeTypes)`, and 
```

P0009/P0009.md
```diff
@@ -1407,26 +1418,39 @@ template<size_t... OtherExtents>
...
-    * [4.1]{.pnum} `(is_convertible_v<SizeTypes, size_type> && ...)` is `true`, and
+    * [5.1]{.pnum} `(is_convertible_v<SizeTypes, size_type> && ...)` is `true`, and
     
-    * [4.2]{.pnum} `sizeof...(SizeTypes) == rank_dynamic()` is `true`.
+    * [5.2]{.pnum} `sizeof...(SizeTypes)` is `dynamic_rank()` or `rank()`.
 
-* [5]{.pnum} *Effects:*  Initializes `dynamic_extents_` with `dynamic...`.
+* [5]{.pnum} *Preconditions:* If `sizeof...(SizeTypes) == rank()` is `true`,
+             $r$ is smaller than `sizeof()...SizeTypes)`, and 
+             $E_r$ is a _static extent_, than
```
mhoemmen, line 1432:
```suggestion
             $E_r$ is a _static extent_, then
```

P0009/P0009.md
```diff
@@ -1407,26 +1418,39 @@ template<size_t... OtherExtents>
...
 ```
 
-* [6]{.pnum} *Constraints:* `is_convertible_v<SizeType, size_type>` is `true`.
+* [7]{.pnum} *Constraints:*
+
+   * [7.1]{.pnum} `is_convertible_v<SizeType, size_type>` is `true`, and
+
+   * [7.2]{.pnum} `N` is `rank_dynamic()` or `rank()`.
+
+* [5]{.pnum} *Preconditions:* If `N` is `rank()`, $r$ is smaller than `N` and $E_r$ is a _static extent_, than
```
mhoemmen, line 1449:
```suggestion
* [5]{.pnum} *Preconditions:* If `N` is `rank()`, $r$ is smaller than `N`, and $E_r$ is a _static extent_, then
```

mhoemmen (changes requested):
Please see change suggestions; thanks!

P0009/P0009.md
crtrott: i think that was a general thing that you don't say "then" for the returns statement.

P0009/P0009.md
crtrott: done

P0009/P0009.md
crtrott: done

P0009/P0009.md
crtrott: done

P0009/P0009.md
crtrott: done

P0009/P0009.md
crtrott: done

P0009/P0009.md
crtrott: done

P0009/P0009.md
crtrott: done

P0009/P0009.md
```diff
@@ -1407,26 +1418,39 @@ template<size_t... OtherExtents>
...
+* [5]{.pnum} *Constraints:*
     
-    * [4.1]{.pnum} `(is_convertible_v<SizeTypes, size_type> && ...)` is `true`, and
+    * [5.1]{.pnum} `(is_convertible_v<SizeTypes, size_type> && ...)` is `true`, and
     
-    * [4.2]{.pnum} `sizeof...(SizeTypes) == rank_dynamic()` is `true`.
+    * [5.2]{.pnum} `sizeof...(SizeTypes)` equals either `dynamic_rank()` or `rank()`.
 
-* [5]{.pnum} *Effects:*  Initializes `dynamic_extents_` with `dynamic...`.
+* [5]{.pnum} *Preconditions:* If `sizeof...(SizeTypes) == rank()` is `true`,
```
tomaszkam, line 1430: This preconditions needs to be always checked. Imagine E = extends<dyn, 1, 2, dyn, 3, 4>. Then E e(1, 1, 2, 4) requires second and third parameter to be equal to static extends. I.e. static_extends may appaer before dynamic extend.

P0009/P0009.md
```diff
@@ -1407,26 +1418,39 @@ template<size_t... OtherExtents>
...
 ```
 
-* [4]{.pnum} *Constraints:*
+* [5]{.pnum} *Constraints:*
     
-    * [4.1]{.pnum} `(is_convertible_v<SizeTypes, size_type> && ...)` is `true`, and
+    * [5.1]{.pnum} `(is_convertible_v<SizeTypes, size_type> && ...)` is `true`, and
     
-    * [4.2]{.pnum} `sizeof...(SizeTypes) == rank_dynamic()` is `true`.
+    * [5.2]{.pnum} `sizeof...(SizeTypes)` equals either `dynamic_rank()` or `rank()`.
```
tomaszkam, line 1428: Why do not allow any value in range [dynamic_rank(), rank()). Any further extends would need to be equal to the corresponding static extend. But this is already the case for any static rank, present before last dynamic extend.

P0009/P0009.md
crtrott: Hey, so we had an author meeting and there was pretty strong opposition to the "leave trailing arguments off" option. So we only allow either provide only the dynamic extends or provide all. If you provide only the dynamic extents there is nothing to check.

crtrott: Oh and we can certainly see if LEWG on the 22nd wants to vote on requiring us to do one of the other options) (hopefully while saying it can be forwarded with that change, if the vote outcome is to do something else than what we do now).

P0009/P0009.md
tomaszkam: This was answered by above.

P0009/P0009.md
tomaszkam: Ok, thanks. This also answers my other comment.

P0009/P0009.md
crtrott: This was the majority vote from the authors, only allowing either `rank_dynamic` or `rank` (just noticed I have a word swap there for `rank_dynamic`). If we did something different I think it had to be between "arguments up to last dynamic rank" and rank. It wouldn't be between "rank_dynamic" and "rank" in that case. That said I did do a test implementation of all options.

tomaszkam: The constructor of `mdspan` (and `extends`) should be conditionally explicit if `sizeof...(Args) > rank_dynamic()` or `N > rank_dynamic()`, i.e. when there is possibility of UB, to match span (ptr, size) constructor.
Note that `N > rank_dynamic()` is same as `N == rank() && rank_dynamic() != rank()` with your precondition. Could be also spelled as `N != rank_dynamic()`

crtrott: yeah it should be. Doing that now.

crtrott: Thanks for looking at this!

mhoemmen (approved):

PR #186, opened 2021-12-15
P1673R6 changes

mhoemmen:
Changes for P1673R6.

### DIFF
--- a/D1673/P1673.bs
+++ b/D1673/P1673.bs
@@ -1,8 +1,8 @@
 <pre class='metadata'>
-Title: D1673R4: A free function linear algebra interface based on the BLAS
-Shortname: D1673R4
-Status: iso/D
-Level: 4
+Title: P1673R5: A free function linear algebra interface based on the BLAS
+Shortname: P1673R5
+Status: iso/P
+Level: 5
 Group: WG21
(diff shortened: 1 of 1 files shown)


crtrott:
I don't think we should remove the overloads without alpha, its not ambiguous at all if you use the correct requires clauses.

crtrott (changes requested):
I don't think we should remove the overloads - at least not without talking to all the other authors. There is no ambiguity here if you use the correct preconditions (i.e. C++20 requires clauses).

mhoemmen:
> I don't think we should remove the overloads - at least not without talking to all the other authors. There is no ambiguity here if you use the correct preconditions (i.e. C++20 requires clauses).

I'll put them back in the next version.  That was a bit of a hasty decision.  On the other hand, the ambiguity would be challenging to resolve without some complicated `requires` clauses, and I'm not sure if LEWG would appreciate the lack of precedent. 

In any case, I submitted this as R6, so please do merge it.  I can file an issue to bring back the overloads, if we can figure out how to constrain them correctly.

mhoemmen:
Superseded by PR https://github.com/ORNL/cpp-proposals-pub/pull/222 .  Thanks!

mhoemmen:
Merge of https://github.com/ORNL/cpp-proposals-pub/pull/222 means we can delete this branch.

PR #188, opened 2021-12-30
P0009: `operator==` for `layout_stride` should consider strides

MattStephanson:
I guess this is the intended behavior, although the layout mapping requirements table doesn't show `op==`, so it's not clear what the semantics are supposed to be in general.


crtrott (approved):
I will add the changelog entry after, to not conflict with the other change we made to mdspan rank returning size_t.

PR #190, opened 2022-01-13
P0009: misc. wording suggestions

MattStephanson:
I mentioned to @crtrott by private email that I was working on an `mdspan` implementation. I'm still implmenting `submdspan`, but I think I've now read all the wording closely enough to offer a single set of wording suggestions. From the changelog:

- Consistently use "`<`_expr_`>` is `true`" instead of "`<`*expr*`>` is true".
- In the layout mappings' `operator()` *Effects* clauses, use only `index_sequence`and fix syntax error in `stride(P())`.
- Replace various unary folds with binary to handle `extents<>`.
- Consistently use `Extents::rank()` in layout mapping wording.
- Typo in `mdspan::mapping_type`: `template mapping_type` should be `template mapping`.
- `mdspan`'s `array` constructor calls `Extents`'s `array` constructor.
- Clarify the value of `static_stride` for `submdspan`'s result.

I'm happy to discuss or clarify any of these, or remove any that you'd rather not change.

### DIFF
--- a/P0009/P0009.md
+++ b/P0009/P0009.md
@@ -40,6 +40,13 @@ toc: true
 - `mdspan::rank[_dynamic]` returns `size_t`
 - fix comparison operator for `layout_stride` to take strides into account.
 - fix `layout_stride` mapping `required_span_size`
+- Consistently use "`<`_expr_`>` is `true`" instead of "`<`*expr*`>` is true".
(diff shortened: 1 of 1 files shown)


crtrott (approved):

crtrott:
Thanks I went carefully through all the changes and those are good

Issue #205, opened 2022-03-04
P0009 LWG Review 03/04/2022


crtrott:
Label naming: https://github.com/cplusplus/draft/wiki/Specification-Style-Guidelines#labels

### mdspan.terms

- par 2: remove ... from idx
- par 3: remove $ signs 

### Extents synopsis
- extents::rank_dynamic call dynamic-index instead of fold
- change comment: Constructors and assignment to Constructors, label change to mdspan.extents.ctors
- mdspan.extents.compare -> mdspan.extents.cmp
- paragraph 3-5 being one, leave 6
- rev-prod-of-extents misses argument name e and i in synopsis and definition
- Preconditions missing dot: 7 9 11 13
- par: 12 and 14 comma after condition, and "the product" 
- Separate heading for exposition only members where the line is
- make fwd/rev-prod-of-extents public member functions
  - accept rank-type
  - add note: these functions are only invoked in situations where this can't overflow
 

### mdspan.extents.cons
- converting constructor: 
  - put the explicit condition into the see below (but not in the synopsis)  
  - For each E_r that is not dynamic index, E_r is equal to other.extent(r).

- par 5.3: dynamic extents should not be in italics

Thought experiment but don't do this:
template<class...SizeTypes
  explicit constexpr extents(SizeTypes... exts)
- change to have two constructors with dynamic and rank size_types
- try this in implementation

- qualify move/forward with std:: everywhere

- par 7: remove italics from static extent

- par 9.1 "is" true instead of "equals" true
- par 9.3 (N=rank_dynamic || N==rank())  is `true`

- par 11.1 element wise wording like 11.2
  - also check that in `layout_stride`

- just do `explicit extents(Integrals...) -> see below`

### mdspan.extents.obs
- put back in "static" before static_extent

### mdspan.layout
- Fix numbering
- par 1.6: the integer packs i and j are ...
- par 1.8: dr is a pack of (possibly const) pack of integers 
  - make dr a d_r like E_r

crtrott:
Things we did as part of this:

commit: 7366664af806656

- add size of a multidimensional index space definition, used later as part of preconditions for example for layouts
- add REQUIRED-SPAN-SIZE exposition only macro for layout_stride

- Fixes to requirement lists:
  - changed "Value" to "Returns"
  - Removed "type" preconditions and reformulated results
  -

PR #206, opened 2022-03-08
Lwg review 03 04 22

crtrott:


crtrott:
ah this has the preliminary changes for P2553 from PR #204 
But I don't want to rebase because #205 has all the sha's in it for the changes to P0009.

I anyway will need to rewrite P2553 completely to rebase it on top of P0009 so I think its ok to merge this for now and then fix the comments we got in #204 separately if still needed after rebasing.

PR #228, opened 2022-05-10
add/fix things related to tensors and BLAS

jeffhammond:
0. infintely is spelled wrong but also wrong in meaning.  there are not infinitely more possibilities for tensors, merely combinatorially many more.
1. tensors over BLAS is not optimal although it is common.
2. added citations related to tensors and BLAS.
3. GotoBLAS is obsolete.  added OpenBLAS instead.
4. added BLIS and citation.

### DIFF
--- a/D1673/blas_interface.md
+++ b/D1673/blas_interface.md
@@ -329,7 +329,7 @@ in support of adding linear algebra to the C++ Standard Library.
   and Cerebras' [Wafer Scale
   Engine](https://www.cerebras.net/product/#chip).  Several large
   computer system vendors offer optimized linear algebra libraries
-  based on or closely resembling the BLAS; these include AMD's BLIS,
+  that implement the BLAS API; these include AMD's fork of BLIS,
   ARM's Performance Libraries, Cray's LibSci, Intel's Math Kernel
   Library (MKL), IBM's Engineering and Scientific Subroutine Library
   (ESSL), and NVIDIA's cuBLAS.


D1673/blas_interface.md
```diff
@@ -834,12 +835,16 @@ to some multi-indices in the Cartesian product of extents.
 ### Tensors
 
 We exclude tensors from this proposal, for the following reasons.
-First, tensor libraries naturally build on optimized dense linear
+First, tensor libraries often build on optimized dense linear
 algebra libraries like the BLAS, so a linear algebra library is a good
```
mhoemmen, line 839:
```suggestion
algebra libraries like the BLAS.  Thus, a linear algebra library is a good
```

D1673/blas_interface.md
```diff
@@ -1597,7 +1602,7 @@ pioneering efforts and history lessons.
   SLATE Working Notes, Innovative Computing Laboratory, University of
   Tennessee Knoxville, Feb. 2018.
 
-* K. Goto and R. A. van de Geijn, "Anatomy of high-performance matrix
+* K. Goto and R. A. van de Geijn, ["Anatomy of high-performance matrix
```
mhoemmen, line 1605: Good catch!

mhoemmen (changes requested):
Thanks @jeffhammond for the fixes!  We should clarify that only P1673.bs needs to be edited.  I'm not sure why we haven't deleted the old file `blas_interface.md` yet.  If you have the time, it would be excellent to delete it in this PR.

mhoemmen:
I added PR https://github.com/ORNL/cpp-proposals-pub/pull/231, which deletes the old `blas_interface.md` file.  I can also check P1674 to see whether we could apply the changes there.

mhoemmen:
( @crtrott @nliber @dalg24 )

PR #238, opened 2022-05-12
P1673: Fix wording for user-defined complex types

### LINKED ISSUE #236: P1673: Fix wording to permit user-defined complex number types
## Summary

Thanks to @PhilMiller for an excellent discussion about this!  I've written some text to add to the introduction, but we also need normative wording changes.  The essence is to replace uses of `conj` in the wording with an exposition-only function _`conj-if-needed`_, defined as follows:

```
template<class T>
T <it>conj-if-needed</it>(const T& t)
{
    using std::conj;
    return conj(t);
}

template<class T> requires is_arithmetic_v<T>
T <it>conj-if-needed</it>(const T& t)
{
    return t;
}
```

If we needed actual words for this, we could borrow them from `iter_swap`'s wording.  Here follows a draft of the text.

## Support for user-defined complex number types

The C++ Standard currently supports `complex<T>` for T = `float`, `double`, and `long double`.  Users who want complex numbers of other types (such as integers or custom "bigfloat" types), or who have different alignment requirements for complex numbers, need to make their own complex number class (or class template).  We want P1673 to "do the right thing" for these custom complex numbers, without requiring any more number traits machinery than the Standard already has.  This means relying on ADL to pick up overloads of `conj`, and not defining any new traits (such as `is_complex`).  The idiom of `using std::conj; return conj(x);`, analogous to how `iter_swap` accesses `swap`, can manage this.  Here is an example with a user-defined complex number class template `MyComplex`.

```c++
#include <cassert>
#include <complex>

namespace MyNamespace {

template<class Real>
struct MyComplex {
    Real real;
    Real imag;
};

template<class Real>
MyComplex<Real> conj(const MyComplex<Real>& z) {
    return {z.real, -z.imag};
}

} // MyNamespace

int main() {
    using std::conj;

    MyNamespace::MyComplex<int> z{3, 4};
    const auto z_conj = conj(z);
    assert(z_conj.real == 3);
    assert(z_conj.imag == -4);

    std::complex<double> w{5.0, 6.0};
    const auto w_conj = conj(w);
    assert(w_conj.real() == 5.0);
    assert(w_conj.imag() == -6.0);
    return 0;
}
```

Regarding alignment: The Standard explicitly specifies that `complex<T>` has the same alignment as `T[2]`.  That's not so friendly for GPUs and SIMD.  `Kokkos::complex` also exists because a few years ago, we had to bless its member functions with `__host__` `__device__` .)  (I wrote `Kokkos::complex` back then, so that we could build Trilinos with complex number support on GPUs.)

If we define `accessor_conjugate` to do `using std::conj; return conj(x);`, then we can support custom complex class templates.  Users can define `conj` for their custom complex type outside the `std` namespace, and this approach will pick it up, analogous to how `iter_swap` can pick up a custom `swap` overload.  However, this will do the wrong thing for built-in (non-complex) arithmetic types like `double` or `int`.  This is because `std::conj` is defined always to return a specialization of `complex`, even if the input is (say) `double`.

One could argue that applying `conjugated` to a view of real numbers *should* be ill formed.  However, users may want to use `conjugate_transposed` to write generic code that works for both real and complex numbers.  Matlab takes this approach, by defining the `'` (single quote) operator to take the conjugate transpose if its argument is complex, and the transpose if its argument is real.  BLAS also supports this to some extent, by having its real routines like `DGEMM` interpret `TRANSA='C'` as the transpose.

We can support `conjugated` (and thus `conjugated_transposed`) for all types as follows.

1. For arithmetic types, define `accessor_conjugate` specially to return just the reference.  Don't use `conj` at all.

2. For all other types (including specializations of `complex`), define `accessor_conjugate` to use `using std::conj; return conj(x);`.

We could express this through an exposition-only function _`conj-if-needed`_:

```
template<class T>
T <it>conj-if-needed</it>(const T& t)
{
    using std::conj;
    return conj(t);
}

template<class T> requires is_arithmetic_v<T>
T <it>conj-if-needed</it>(const T& t)
{
    return t;
}
```

Even though it's possible to write "generic" (real or complex) linear algebra code using `conjugate_transposed`, it's still important to distinguish between symmetric and Hermitian matrix algorithms.  This is because symmetric does *not* mean the same thing as Hermitian for matrices of complex numbers.  Complex symmetric matrices are useful in practice, for example when modeling damped vibrations (Craven 1969).

B. D. Craven, ["Complex symmetric matrices"](https://doi.org/10.1017/S1446788700007588), *Journal of the Australian Mathematical Society*, Vol. 10, No. 3-4, Nov. 1969, pp. 341--354.

Credit to Phil Miller (`phil.miller@intensecomputing.com`) for pointing out a wording issue that would have hindered use of user-defined complex types.

mhoemmen:
@crtrott @dalg24 @nliber 

* Add intro wording explaining the design
* Make sure `accessor_conjugate` and `conjugated` work correctly for user-defined complex types
* Introduce _`conj-if-needed`_ to simplify wording
* Resolve vagueness in return type of `conjugated`
* Add acknowledgments for Yu You and Phil Miller
* Add R8 submission date
* Update my e-mail address

Fixes #236.
Addresses part of https://github.com/kokkos/stdBLAS/issues/189.

### DIFF
--- a/D1673/P1673.bs
+++ b/D1673/P1673.bs
@@ -2237,12 +2237,11 @@ This is a good way to build a noncommutative semiring for testing.
     gives implementers the freedom to make QoI enhancements
     without risking correctness.
 
-    "Value type constraints do not suffice to describe algorithm behavior" section above,
-
 * We think describing algorithms' behavior and implementation freedom
     is more useful than mathematical concepts like "ring."
     For example, we permit implementations to reorder sums,
     but this does not mean that they assume sums are associative.
+    This is why we do not define a hierarchy of number concepts.
 
 * We categorize different ways that implementers might like to change algorithms,
     list categories we exclude and categories we permit,
@@ -2258,8 +2257,10 @@ This is a good way to build a noncommutative semiring for testing.
 
 Summary:
 
-1. Consider generalizing function parameters to take any type that implements the
-    `get_mdspan` customization point, including `mdarray`.
+1. Consider generalizing function parameters to take any type
+    that implements a customization point
+    for getting an `mdspan` viewing its elements.
+    This includes `mdarray` (see P1684).
 
 2. Add batched linear algebra overloads.
 
--- a/D1673/P1673.bs
+++ b/D1673/P1673.bs
@@ -260,6 +260,8 @@ Date: 2022-05-15
 
 * Revision 8 (electronic), to be submitted 2022-05-15
 
+    * Resolve vagueness in `const`ness of return type of `transposed`
+
     * Resolve vagueness in `const`ness of return type of `scaled`,
         and make its element type the type of the product,
         rather than forcing it back to the input `mdspan`'s element type
@@ -5671,11 +5673,18 @@ transposed(
 
 Let `ReturnExtents` name the type `transpose_extents_t<Extents>`.
 Let `R` name the type
-`mdspan<ReturnElementType, ReturnExtents, ReturnLayout, Accessor>`,
+`mdspan<ReturnElementType, ReturnExtents, ReturnLayout, ReturnAccessor>`,
 where
 
-* `ReturnElementType` is either `ElementType` or
-    `const ElementType`; and
+* `ReturnElementType` is:
+
+    * if `Accessor` is `default_accessor<ElementType>`,
+        then `add_const_t<ElementType>`;
+
+    * else, `ElementType`; *[Note:* For general accessors,
+        the `const ElementType` version of the accessor may not exist;
+        Even if it does, `transposed` has no way to deduce it generically.
+        --*end note]*
 
 * `ReturnLayout` is:
 
@@ -5695,7 +5704,14 @@ where
         for some `NestedLayout`, then
         either `NestedLayout` or `layout_transpose<Layout>`,
 
-    * else `layout_transpose<Layout>`.
+    * else `layout_transpose<Layout>`;
+
+* and, `ReturnAccessor` is:
+
+    * if `Accessor` is `default_accessor<ElementType>`,
+        then `default_accessor<add_const_t<ElementType>>`;
+
+    * else, `Accessor`.
 
 * *Effects:*
 
--- a/D1673/P1673.bs
+++ b/D1673/P1673.bs
@@ -260,6 +260,9 @@ Date: 2022-05-15
 
 * Revision 8 (electronic), to be submitted 2022-05-15
 
+    * Optimize `transposed` for all known layouts,
+        so as to avoid use of `layout_transpose` when not needed
+
     * Fix matrix extents in constraints and mandates
         for `symmetric_matrix_rank_k_update` and `hermitian_matrix_rank_k_update`
         (thanks to Mikołaj Zuzek for reporting the issue)
@@ -5697,11 +5700,7 @@ The `transposed` function takes a rank-2 `mdspan`
 representing a matrix, and returns a new read-only `mdspan`
 representing the transpose of the input matrix.  The input matrix's
 data are not modified, and the returned `mdspan` accesses the
-input matrix's data in place.  If the input `mdspan`'s layout is
-already `layout_transpose<L>` for some layout `L`, then the returned
-`mdspan` has layout `L`.  Otherwise, the returned `mdspan`
-has layout `layout_transpose<L>`, where `L` is the input
-`mdspan`'s layout.
+input matrix's data in place.
 
 ```c++
 template<class ElementType,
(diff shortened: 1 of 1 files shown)


D1673/P1673.bs
```diff
@@ -5008,22 +5273,15 @@ auto operator* (const T1 x, const conjugated_scalar<Reference, Element> y) {
 
 * *Preconditions:* `Reference` shall be *Cpp17CopyConstructible*.
 
-* *Constraints:* The expression `conj(val)` is well formed and is convertible to
-    `ElementType`.  *[Note:* This implies that `ElementType` is
-    `complex<R>` for some type `R`.  --*end note]*
+* *Constraints:* The expression _`conj-if-needed`_`(val)` is well formed
+    and is convertible to `ElementType`.
```
PhilMiller, line 5277: I think there's a spelling mismatch here of `Element` vs `ElementType`

D1673/P1673.bs
```diff
@@ -3947,6 +4188,19 @@ void triangular_matrix_matrix_right_solve(
   Triangle t,
   DiagonalStorage d,
   inout_matrix_t B);
+
+template<class T>
+T <it>conj-if-needed</it>(const T& t) // <it>exposition only</it>
+{
+    using std::conj;
+    return conj(t);
+}
```
PhilMiller, line 4197: I'm stretching here, but should `std::blas` support types whose conjugate is not the same type? As written, this implicitly says "no".

I'm struggling to come up with good examples other than dependent-type trickery (e.g. polar representations whose types limit their `theta`), but that feels like my mathematical ignorance, not fundamental.

It definitely seems reasonable to me to say that computational efficiency demands this equivalence, though.

D1673/P1673.bs
```diff
@@ -7479,8 +7758,8 @@ impossible to express the update A = A - x x^H otherwise.
 For `i,j` in the domain of `A`
 and in the triangle of `A` specified by `t`,
 the mathematical expression for the algorithm is
-`A[i,j] += x[i] * conj(x[j])` for overloads without an `alpha` parameter, and
-`A[i,j] += alpha * x[i] * conj(x[j])` for overloads without an `alpha` parameter.
+`A[i,j] += x[i] * `_`conj-if-needed`_`(x[j])` for overloads without an `alpha` parameter, and
+`A[i,j] += alpha * x[i] * `_`conj-if-needed`_`(x[j])` for overloads without an `alpha` parameter.
```
PhilMiller, line 7762: Incidentally, I notice what looks like a copy-paste error here. Should this read 
```suggestion
`A[i,j] += alpha * x[i] * `_`conj-if-needed`_`(x[j])` for overloads with an `alpha` parameter.
```

D1673/P1673.bs
mhoemmen: > I'm stretching here, but should std::blas support types whose conjugate is not the same type? As written, this implicitly says "no".

For all arithmetic types, `conj(x)` is `complex<double>`.  The point of _`conj-if-needed`_ is to fix that weird thing, so that (for the sake of computational efficiency), `conjugated` preserves the value type of its input.

For (real, imaginary) complex number representations, it would be weird if conjugation behaved differently than unary minus.  For (magnitude, argument) representations, conjugation just rotates by 180 degrees.  In either case, I find it hard to imagine why `conj` should change the type.

D1673/P1673.bs
mhoemmen: Good catch!  I'll fix this in #242.  Thanks!

D1673/P1673.bs
mhoemmen: Good catch!  I'll fix this in #242.  Thanks!

D1673/P1673.bs
mhoemmen: That being said, the easy fix for types like this would be for _`conj-if-needed`_ to return `auto` instead of `T`.  Thanks for the suggestion!  I'll make this change in PR #242.

mhoemmen: I've made the change in PR #242 -- some `accessor_conjugate` wording had to change too.  Thanks!

PhilMiller:
I think there may still be issues with the approach taken to `conj-if-needed` and `use std::conj; conj(x);`.

Suppose a user-defined `class MyReal` follows the same reasoning for implementing `conj(MyReal)` as `std::conj(double)`, and defines it as `MyComplex conj(MyReal r)`. We don't want to end up using that in a conjugation-generic algorithm that's applied to real input and expects to produce real output. I see this as a pretty fundamental, if perhaps inadvertent, customization point.

I don't know whether the reasoning that got us `std::complex<double> std::conj(double)` was a mistake, or was based on a different set of premises than this effort, but I think that privileging the language's primitive arithmetic types over anything users may define in subsequent work would be an unfortunate path.

A type trait of some sort may not be essential to addressing this, but it does seem like the cleanest path. It could be as broad as a notional `std::is_real_valued<T>`, or as narrow as `std::blas::should_conjugate<T>`. I'll take a look at other potential approaches that may feel like a lower footprint.

PhilMiller:
Ugh, I went looking for the rationale for `std::complex<double> std::conj(double)`, and found that [it's inherited from C99. ](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2004/n1568.htm)

mhoemmen:
@PhilMiller I do think this is an important design question to bring to LEWG.  (Hopefully they will be delighted to have a design question in their hands that doesn't call for specialized experience!)

PhilMiller:
Small note for however the conjugation stuff gets resolved - this will likely need to be edited to match:

> 16.6. Conjugated in-place transformation [linalg.conj]

> The conjugated function takes an mdspan x, and returns a new read-only mdspan y with the same domain as x, whose elements are the complex conjugates of the corresponding elements of x. If the element type of x is not `complex<R>` for some R, then y is a read-only view of the elements of x.

This does not support user-defined complex types at all.

mhoemmen:
Thanks Christian!

Issue #266, opened 2022-06-21
P0009: LWG small group review 2022/06/21


mhoemmen:
@crtrott @dalg24 @nliber 

* `extents(const array&)` and `extents(span)` constructors: Rename `SizeType` (local) template parameter to `OtherSizeType`, to prevent shadowing class template parameter

* Change "integral type other than `bool`" to "a signed or unsigned integer type" (to exclude character types)

Mick235711:
> Adding empty() (see PR #262) will require an LEWG paper or NB comment

@mhoemmen @nliber I just finished the paper; the draft is at [here](http://api.csswg.org/bikeshed/?force=1&url=https://raw.githubusercontent.com/Mick235711/wg21-papers/main/draft-empty.bs), pending a paper number.

Really rushed, the wording was copied from R17.

Feel free to make any comments/suggestions!

Original PR now only contains `noexcept` stuff. Technically if we want to be safe we can add that to paper too; however I personally do think they are okay to merge on their own.

crtrott:
@Mick235711 you could potentially just inline return size()!=0 instead of adding the description below?

nliber:
@crtrott There could be a more optimized implementation of `empty()`, since it doesn't have to compute the size.  As specified an implementation has the freedom to do so.

crtrott:
My own notes:

### mdspan.extents
- SizeType -> OtherSizeType for extens constructor from span/array

- par 2.1 - SizeType is a signed or unsigned integer type, and
- par 2.2 - or is representable as a value of type SizeType (check for other instances of this)
  - mdspan.extents.ctor 2.2
  - mdspan.layoutleft.ctor etc. 

- par 9: fix copy paste error
- par 11: size_type -> size_t

### mdspan.layout.reqmts

- par 8. add smaller then `size_t` too

### mdspan.layoutleft.ctor

3: fix all the numbers

### mdspan.layoutstride.ctor

- SizeType -> OtherSizeType
- change back strides() -> array<size_type, rank_>


### mdspan.overview

- SizeType to OtherSizeType for constructors/operator taking span/array
- spaces around `==`  `rank_v<CArray>==1`

### mdspan.oberservers

- repeat and fix precondition for indicies from mappings
  <i>num-cast<i><size_type>(i) is i if remve_cvref_t<decltype(i)> is integer-like type, otehrwise static_cast<size_type>(i)

PR #267, opened 2022-06-21
Lwg small group 0621

crtrott:


crtrott:
Addresses #266

dalg24 (approved):

P0009/P0009.md
```diff
@@ -1390,6 +1392,12 @@ public:
   // [mdspan.extents.helpers], exposition only helpers
   constexpr size_t @_fwd-prod-of-extents_@(rank_type) const noexcept; // @_exposition only_@
   constexpr size_t @_rev-prod-of-extents_@(rank_type) const noexcept; // @_exposition only_@
+  template<class OtherSizeType>
+    requires(!is_integral_v<OtherSizeType>)
+    static constexpr size_type @_index-cast_@(const OtherSizeType& i) noexcept { return static_cast<size_type>(i); } // @_exposition only_@
```
mhoemmen, line 1397: 1. The request was to distinguish between ["integer-like" types](http://eel.is/c++draft/iterator.concept.winc#def:integer-like) and other types.  Integer-like types need not be `integral`.
2. The usual way to do constraints is to start with the most general case, and then constrain.  Otherwise, one may run afoul of subsumption.

I suggest not defining this inline, so that we don't have to worry about concepts and subsumption.  Instead, we could just describe this in words.

```c++
template<class T>
static constexpr auto @_index-cast_@(const T& i) noexcept;
```

*Effects:*

    * If `OtherSizeType` is integer-like (**[iterator.concept.winc]**), then equivalent to `return i;`;

    * otherwise, equivalent to `return static_cast<size_type>(i);`.

mhoemmen (changes requested):

P0009/P0009.md
mhoemmen: (Integer-class-but-not-integral would be covered by is_convertible_v<Indices, size_type> anyway, so we can actually say integral here.)

PR #268, opened 2022-06-23
Update P0009.md

mhoemmen:
@crtrott @dalg24 @nliber 

For justification, please see today's (2022/06/22) discussion with Walker, and the following Compiler Explorer example:

https://godbolt.org/z/vrqhacqxo

### DIFF
--- a/P0009/P0009.md
+++ b/P0009/P0009.md
@@ -43,6 +43,7 @@ Special thanks to Tomasz Kaminski for invaluable help in preparing this paper fo
 - `submdspan` was moved into its own paper, allowing mdspan to go forward even if there is not enough time to review `submdspan`
 - Merge P2553 (SizeType as a template parameter for `extents`).
 - Merge P2554 and fix merge conflict with P2553
+- Fix `mdspan` CArray and Pointer deduction guides so they work with non-lvalues
 
 ## P0009r17: 2022-04 Mailing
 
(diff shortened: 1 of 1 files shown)


mhoemmen:
I tried this out in the actual implementation, and it doesn't work: https://godbolt.org/z/E4G9xfehc

Search for `NEW_MDSPAN_DEDUCTION_GUIDES`.  Neither the current code nor the new code correctly deduces from `char[3]`.

mhoemmen:
The implementation's deduction guides differ a bit from P0009's.  I changed the code and still didn't get it to work:

https://godbolt.org/z/aPr3jTPTz

mhoemmen:
Christian and I worked it out tonight.

crtrott:
https://godbolt.org/z/crcEzYn13

I think we have it almost right just need to take const Pointer& in the deduction guide. 

Note we do NOT want to deduce from std::move(carray) because thats dangling immediately (we never store a carray, we get a ptr to the array ...). We also don't want to deduce anything from multi-dimensional carrays, since folks would likely expect a 2d-mdspan, but we can't make a 2d mdspan because of the aliasing rules.

crtrott:
FYI: the implementation (which as of right now still needs an extra define in order to enable P2554 behavior) actually had it as const Pointer& and thus it did work. I also checked, and we did have most of the tests we would think we should have (again required the define being externally set during build). Closing this.

crtrott:
https://godbolt.org/z/6ejvY5Yxz

That is the link with the current mdspan impl, which works as expected.

PR #283, opened 2022-07-09
Mdarray updates

crtrott:


crtrott:
Still missing deduction guide and constructor for mdspan.

PR #285, opened 2022-07-15
Mdarray updates 3

crtrott:


P1684-mdarray/P1684.md
```diff
@@ -298,11 +308,27 @@ class mdarray {
 };
 ```
 
+Moreover, containers such as `vector` and container adapters like `queue` have a significant number of other constructors which copy the data on input, such as constructors taking ranges, input iterator pairs, and initializer lists.
+The R2 version of this paper had a number of those, but in combination with allocator arguments it lead to an explosion of constructors.
+A draft version seen by LEWG on 2022-07-12 sported 41 constructors. 
+However almost all these constructors would simply forward arguments to teh constructor of the container owned by `mdarray`.
```
dalg24, line 314:
```suggestion
However almost all these constructors would simply forward arguments to the constructor of the container owned by `mdarray`.
```

P1684-mdarray/P1684.md
```diff
@@ -298,11 +308,27 @@ class mdarray {
...
+LEWG provided feedback that we should consider reducing the amount of constructors, since at a minimum one could inline construct the `container` itself, and move it into the `mdarray` upon construction.
+
+```c++
+// with convenience constructors:
+mdarray a(first, last, N, M);
+
+// without
+mdarray a(std::move(vector(first, last)), N, M);
```
dalg24, line 322: Why is it not just `mdarray a(vector(first, last), N, M);`?

PR #296, opened 2022-09-09
[REVIEW] layout_left_padded & layout_right_padded

mhoemmen:
1. Add new mdspan layouts `layout_left_padded` and `layout_right_padded`
2. Add support for those layouts in submdspan
3. Change submdspan of `layout_left` resp. `layout_right` to produce `layout_left_padded` resp. `layout_right_padded` if possible

### DIFF
--- a/layout_padded/P2642R0.html
+++ b/layout_padded/P2642R0.html
@@ -2,7 +2,7 @@
  <head>
   <meta content="text/html; charset=utf-8" http-equiv="Content-Type">
   <meta content="width=device-width, initial-scale=1, shrink-to-fit=no" name="viewport">
-  <title>DXXXXR0: Padded mdspan layouts</title>
+  <title>P2642R0: Padded mdspan layouts</title>
 <style data-fill-with="stylesheet">/******************************************************************************
  *                   Style sheet for the W3C specifications                   *
  *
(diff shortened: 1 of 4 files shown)


crtrott:
Just a few random comments for now:

unpadded_extents_type: maybe rename to unpadded-extent-type (i.e. singular and dashes)

I don't want to constraint the constructor which just takes extents to static padding: just make the padding essentially 0 (i.e. no padding). That would make the layout always constructible from extents, which you could for example use with the mdarray constructor which takes just extents. I think we will end up with lots of codes which is gonna be constrained to taking mdspans constructible from extents only. 

The constructor which takes both extents and padding value:
Why a default value of 0?? Doesn't that mean that 
layout_left_padded<dyn>::mapping<extents<size_t,dyn,dyn>>(extents(2,2)); results in a mapping where the inner mapping has  map.extent(0)==0?

So I think: make the first constructor available also with dynamic_extent padding, which then gets set to the extent(0) of the inner mapping, take the default value away from the second constructor, and make it available indpendent of whether you have dynamic_extent padding or static. In the static case you have a precondition for matching the thing. 

Generally we need mandates/preconditions for the padding to be larger equal the extent value (mandates when extent and padding are static).

Ok too tired to do more stuff.

mhoemmen:
TODO:

- [ ] Do not constrain constructor that just takes extents to static padding; just use no padding in that case.
- [ ] Remove default padding value from the constructor taking both extents and padding value.  In static case, add precondition for matching `padding_stride`.

mhoemmen:
@crtrott @dalg24 @nliber This is the version I submitted on 2022/09/15.  As such, it would be excellent if you could merge this without changes first; thanks!

PR #313, opened 2022-12-30
P2642: R2 update (2022/12/28 - 2022/12/29)

mhoemmen:
P2642 R2 updates started 2022/12/28 and finished 2022/12/29.  @crtrott , please review; thanks!

This PR does NOT attempt to reformat from Bikeshed to Pandoc.  It only makes changes to content (mostly wording).

### DIFF
--- a/layout_padded/layout_padded.bs
+++ b/layout_padded/layout_padded.bs
@@ -1042,7 +1042,7 @@ let `P_left` denote the parameter pack of elements of type `size_t` such that:
 * if `extents_type::rank()` equals zero or one,
     `P_left` is the empty parameter pack;
 
-* else, `P_left` is the parameter pack of size `extents_type::rank() - 2`
+* else, `P_left` is the parameter pack of size `extents_type::rank() - 1`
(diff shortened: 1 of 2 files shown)


layout_padded/layout_padded.bs
```diff
@@ -803,19 +840,53 @@ template<size_t other_padding_stride, class OtherExtents>
...
+
+The above Mandates clause and some subclauses of [mdspan] below
+mandate that the least multiple of a `constexpr` `size_t` value $p$
+(such as `other_padding_stride` above)
+greater than or equal to some other nonnegative `constexpr` value $e$ of integral type `U`
+is representable as a value of some (possibly different) integral type `V`,
+where $e$ is always representable as a value of type `V`.
+We explain here how to express this condition as a C++ constant expression.
+
+If $p$ is zero, then $e$ must be zero.  (FIXME)
```
mhoemmen, line 861: We'll need to address this issue that "least multiple of $p$ greater than to equal to $e$" doesn't work if $p$ is zero.

mhoemmen: Fixed in my most recent commit.

crtrott (approved):

PR #314, opened 2023-01-03
P1673: Start on R11 changes and Pandoc-ification

mhoemmen:
@crtrott @dalg24 

* Make document build with Pandoc and produce reasonable output.  (Pandoc conversion is ongoing; its completion can be deferred to future pull requests.)

* Replace name-based mdspan requirements with exposition-only concepts.

* Experimentally simplify `*_matrix_vector_product` wording (please review that part with care).

### DIFF
--- a/D1673/P1673.md
+++ b/D1673/P1673.md
@@ -7573,13 +7573,7 @@ void matrix_vector_product(ExecutionPolicy&& exec,
                            out_vector_t y);
 ```
 
-For `i` in the domain of `y` and `N` equal to `A.extent(1)`,
-the mathematical expression for the algorithm is
-`y[i] =` the sum of `A[i,j] * x[j]`
-for all `j` such that `i,j` is in the domain of `A`.
-
-* *Effects:* Assigns to the elements of `y` the product of the matrix
-    `A` with the vector `x`.
+* *Effects:* Computes $y = A x$.
 
 [*Example:*
 ```c++
@@ -7639,13 +7633,7 @@ void matrix_vector_product(ExecutionPolicy&& exec,
                            out_vector_t z);
 ```
 
-For `i` in the domain of `z` and `N` equal to `A.extent(1)`,
-the mathematical expression for the algorithm is
-`z[i] = y[i] + A[i,j] * x[j]`
-for all `j` such that `i,j` is in the domain of `A`.
-
-* *Effects:* Assigns to the elements of `z` the elementwise sum of
-    `y`, and the product of the matrix `A` with the vector `x`.
+* *Effects:* Computes $z = y + A x$.
 
 #### Symmetric matrix-vector product [linalg.algs.blas2.symv]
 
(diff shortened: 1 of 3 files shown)


D1673/P1673.md
```diff
@@ -3469,6 +3506,40 @@ template<class ElementType,
 constexpr auto conjugate_transposed(
   mdspan<ElementType, Extents, Layout, Accessor> a);
 
+// [linalg.concepts], exposition-only concepts
+
+template<class T>
+struct @_is-mdspan_@; // exposition only
```
dalg24, line 3512: Should it say "concept" instead of `struct`?

D1673/P1673.md
```diff
@@ -3484,152 +3555,151 @@ void givens_rotation_setup(const complex<Real>& a,
                            complex<Real>& r);
 
 // [linalg.algs.blas1.givens.rot], apply computed Givens rotation
-template<class inout_vector_1_t,
-         class inout_vector_2_t,
+template<inout_vector InOutVec1,
+         inout_vector InOutVec2,
```
dalg24, line 3559: Can you confirm that is how we want to spell it in code snippets (as compared to `@_inout-vector_@` which I assume would not be rendered properly)?
More particularly with an underscore rather than a dash

D1673/P1673.md
```diff
@@ -6601,103 +6673,110 @@ void test_conjugate_transposed(
...
+_`in-vector`_, _`in-matrix`_, or _`in-object`_ need not be `const`,
+because an mdspan accessor with a nonconst `element_type`
+need not necessarily be convertible to an accessor
+with a const `element_type`. --*end note]*
 
-* `in_matrix*_t` is a rank-2 `mdspan` with a
-    `const` element type.
-    If the algorithm accesses the object, it will do so in read-only fashion.
+If an algorithm in *[linalg.algs]* accesses the elements
+of an _`out-vector`_, `out-matrix`, or `out-object`,
```
dalg24, line 6757: Did you mean to only have `out-vector` in italic, ie not `out-matrix` and `out-object`?

D1673/P1673.md
mhoemmen: Thanks for taking a look at this!  Unlike the other concepts here, `is-mdspan` is a struct (a trait following the usual pattern of inheriting from `true_type` or `false_type`).  On the other hand, this means that it would make sense for me to title this section "exposition-only concepts and traits."

D1673/P1673.md
mhoemmen: All these _`out-vector`_, _`out-matrix`_, and _`out-object`_ need to be italicized.  Thanks for pointing this out!

mhoemmen: Fixed; thanks!

D1673/P1673.md
mhoemmen: This was a formatting bug.  I've fixed this in a comment I pushed today.  Thanks for pointing this out!

D1673/P1673.md
mhoemmen: @dalg24 I changed the title; thanks!

mhoemmen:
@crtrott @dalg24 One issue remains: Some algorithms write to `layout_blas_packed` matrices, which have nonunique layout.  This affects use of _`inout-matrix`_ and possibly also _`out-matrix`_.  A reasonable approach would be for me to add a new exposition-only concept for "_`inout-matrix`_ or mdspan with `layout_blas_packed`," that is, "valid matrix inout argument of `symmetric`, `hermitian`, or `triangular` algorithms."  Ditto for _`out-matrix`_.

mhoemmen:
Thanks Christian! : - D  Regarding my above comment:

All use of _`out-matrix`_ looks correct; none of the algorithms taking _`out-matrix`_ assume that it's symmetric, Hermitian, or triangular.  (Algorithms with `symmetric`, `hermitian`, or `triangular` in their name have an _`in-matrix`_ with that property, but not an _`out-matrix`_.)

mhoemmen:
PR https://github.com/ORNL/cpp-proposals-pub/pull/314 fixes the issue I mentioned above.  Thanks!

PR #329, opened 2023-02-07
P2689R2 - SG1 Issaquah 2023 Changes

nliber:


atomic_accessor/atomic_accessor.md
```diff
@@ -1,6 +1,6 @@
 ---
 title: "Atomic Refs Bounded to Memory Orderings & Atomic Accessors"
-document: P2689R1
+document: P2689R2
```
dalg24, line 3: Does the title need to be updated as well "Bound[ed]"

dalg24: (They are more occurrences that would need fixing)

atomic_accessor/atomic_accessor.md
```diff
@@ -612,12 +654,11 @@ constexpr basic-atomic-accessor(basic-atomic-accessor<OtherElementType, Referenc
...
 constexpr reference access(data_handle_type p, size_t i) const noexcept;
 ```
-[2]{.pnum} *Effects:* Equivalent to `return reference(p + i);`
-
+[2]{.pnum} *Effects:* Equivalent to `return p[i];`
 ```c++
 constexpr data_handle_type offset(data_handle_type p, size_t i) const noexcept;
 ```
-[3]{.pnum} *Effects:* Equivalent to `return p[i];`
+[3]{.pnum} *Effects:* Equivalent to `return reference(p + i);`
```
dalg24, line 681: Ugggg not so sure about the "`reference(...)`" part

atomic_accessor/atomic_accessor.md
nliber: Fixed (in a few places in the prose as well).

atomic_accessor/atomic_accessor.md
nliber: Could you elaborate?  That is the only place the passed in reference type is used.

nliber: Fixed (I think).

dalg24: `offsets` returns `data_handle_type` (aka `ElementType*`) which is not constructible from `reference` (aka `atomic_ref<ElementType>` or alike)

dalg24 (approved):

atomic_accessor/atomic_accessor.md
dalg24: (would have sworn you said "can you elaborate")
anyway looks good now

atomic_accessor/atomic_accessor.md
```diff
@@ -20,6 +20,64 @@ toc: true
 
 # Revision History
 
+## P2689R2 (SG1 Issaquah 2023 discussion)
+
+- Renamed `atomic-ref-bounded` to `atomic.ref.bound`
```
crtrott, line 25: `atomic-ref-bound` ??

crtrott (approved):

PR #333, opened 2023-02-07
Add paper to fix default constructors for all static extents

crtrott:


P2798-fix-default-constructed-mappings/P2798.md
```diff
@@ -0,0 +1,56 @@
...
+
+In subsection 24.7.3.4.5.1 [mdspan.layout.left.overview] add after paragraph 3:
+
+*Mandates:* If `Extents::rank_dynamic() == 0` is `true`, then the size of the multidimensional index space `Extents()` is representable as a value of type `typename Extents::index_type`.
+
+In subsection 24.7.3.4.6.1 [mdspan.layout.right.overview] add after paragraph 3:
+
+*Mandates:* If `Extents::rank_dynamic() == 0` is `true`, then the size of the multidimensional index space `Extents()` is representable as a value of type `typename Extents::index_type`.
+
+In subsection 24.7.3.4.7.1 [mdspan.layout.right.overview] add after paragraph 3:
```
dalg24, line 53:
```suggestion
In subsection 24.7.3.4.7.1 [mdspan.layout.sride.overview] add after paragraph 3:
```

dalg24 (approved):

P2798-fix-default-constructed-mappings/P2798.md
```diff
@@ -0,0 +1,56 @@
...
+layout_left::mapping<extents<int, N, N>> map;
+```
+
+Each individual extent fits into `int` but the theoretical required span size does not.
+
+The non-default constructors already deal with this issue via a preconditions such as:
+
+*Preconditions:* The size of the multidimensional index space `e` is representable as a value of type `index_type` ([basic.fundamental]).
+
+For cases where at least one of the extents is dynamic, the default constructor is also fine, since it will result in an empty index space.
```
mhoemmen, line 39: Is it, though?  What about the following `extents`?

```c++
extents<int, 4'000'000, dynamic_extent, 4'000'000>;
```

Product of the extents is zero mathematically, but will overflow.  Do we need to worry about this case?

I think we can fix this by saying that the forward product of the extents is representable as `index_type`, regardless of whether the extents are dynamic or static.  This would generalize Tomasz's suggestion that we treat `rank_dynamic() == 0` and `rank_dynamic() != 0` in the same way.

P2798-fix-default-constructed-mappings/P2798.md
```diff
@@ -0,0 +1,56 @@
...
+
+For cases where at least one of the extents is dynamic, the default constructor is also fine, since it will result in an empty index space. 
+That leaves only the case for all static extents to be fixed. We propose here to add a mandate to the mapping classes of `layout_left`, `layout_right`, and `layout_stride` that the forward product of extents is representable as `index_type`.
+
+## Proposed Wording
+
+
+In subsection 24.7.3.4.5.1 [mdspan.layout.left.overview] add after paragraph 3:
+
+*Mandates:* If `Extents::rank_dynamic() == 0` is `true`, then the size of the multidimensional index space `Extents()` is representable as a value of type `typename Extents::index_type`.
```
mhoemmen, line 47: Based on the discussion above, we could just leave off the `rank_dynamic() == 0` condition.

```suggestion
*Mandates:* The size of the multidimensional index space `Extents()` is representable as a value of type `typename Extents::index_type`.
```

P2798-fix-default-constructed-mappings/P2798.md
```diff
@@ -0,0 +1,56 @@
...
+
+In subsection 24.7.3.4.6.1 [mdspan.layout.right.overview] add after paragraph 3:
+
+*Mandates:* If `Extents::rank_dynamic() == 0` is `true`, then the size of the multidimensional index space `Extents()` is representable as a value of type `typename Extents::index_type`.
```
mhoemmen, line 51:
```suggestion
*Mandates:* The size of the multidimensional index space `Extents()` is representable as a value of type `typename Extents::index_type`.
```

P2798-fix-default-constructed-mappings/P2798.md
```diff
@@ -0,0 +1,56 @@
...
+
+In subsection 24.7.3.4.7.1 [mdspan.layout.stride.overview] add after paragraph 3:
+
+*Mandates:* If `Extents::rank_dynamic() == 0` is `true`, then the size of the multidimensional index space `Extents()` is representable as a value of type `typename Extents::index_type`.
```
mhoemmen, line 55:
```suggestion
*Mandates:* The size of the multidimensional index space `Extents()` is representable as a value of type `typename Extents::index_type`.
```

P2798-fix-default-constructed-mappings/P2798.md
mhoemmen: I realize this can't be a Mandates if any extents are run time.  We could add a Precondition.

mhoemmen:
```suggestion
*Precondition:* If `Extents::rank_dynamic() != 0` is `true`, then the size of the multidimensional index space `Extents()` is representable as a value of type `typename Extents::index_type`.

*Mandates:* If `Extents::rank_dynamic() == 0` is `true`, then the size of the multidimensional index space `Extents()` is representable as a value of type `typename Extents::index_type`.
```

P2798-fix-default-constructed-mappings/P2798.md
mhoemmen: See above note (this can't be a Mandates if there are any run-time extents).

P2798-fix-default-constructed-mappings/P2798.md
mhoemmen: See above note (this can't be a Mandates if there are any run-time extents).

mhoemmen (approved):
We discussed my concerns over the `rank_dynamic() != 0` case offline, and determined that it was probably out of scope of this paper.  "Don't overflow" might already be an implicit precondition of the forward resp. reverse product of the extents.

P2798-fix-default-constructed-mappings/P2798.md
crtrott: we can't do that. Because the extent for dynamic thing is bad. I.e. its like max size_t value -1 or so. Also This doesn't say how to compute the size of multi-dimensional index space. Its more like the pure math expression.

P2798-fix-default-constructed-mappings/P2798.md
crtrott: you can't do that on a class ...

PR #335, opened 2023-02-09
P1684: Explain why we don't have a way to extract the container from an mdarray

mhoemmen:
Address a question brought up in Issaquah (Feb 2023) LEWG review of P1684R4, by explaining why we don't offer a way to extract the container from an mdarray.

Other changes:

* Update paper number
* Reformat revision history
* Fix `dextents` syntax in examples

@crtrott @dalg24 @nliber


P1684-mdarray/P1684.md
```diff
@@ -644,6 +651,440 @@ The larger or equal is intentional, because the mapping may already be not exhau
...
+std::vector<float>&& f0(std::vector<float>&& storage)
+{
+  mdspan m0{storage.data(), 2, 3, 4};
+  // ... compute with m0 ...
+  return std::move(storage);
+}
+
+std::vector<float>&& f1(std::vector<float>&& storage)
+{
+  mdarray m1{storage.data(), 6, 4};
```
crtrott, line 812:
```suggestion
  mdspan m1{storage.data(), 6, 4};
```

P1684-mdarray/P1684.md
```diff
@@ -644,6 +651,440 @@ The larger or equal is intentional, because the mapping may already be not exhau
...
+
+The disadvantages of this approach are
+
+1. it makes the code more complicated, and
+
+2. it requires control of the container type.
+
+The second point means that users would not be able to call functions
+that use a container with an incompatible allocator type.
+This is somewhat mitigated by mdspan being
```
crtrott, line 767:
```suggestion
This is somewhat mitigated by `mdspan` being
```

P1684-mdarray/P1684.md
```diff
@@ -644,6 +651,440 @@ The larger or equal is intentional, because the mapping may already be not exhau
...
+  return 0;
+}
+```
+
+To elaborate:
+Other languages and libraries have a "reshape" facility
+for reinterpreting an existing array as an array
+with a different (multidimensional) "shape."
+Examples include Fortran's `RESHAPE` intrinsic function.
+The mdarray class cannot be "reshaped in place,"
```
crtrott, line 832:
```suggestion
The `mdarray` class cannot be "reshaped in place,"
```

P1684-mdarray/P1684.md
```diff
@@ -644,6 +651,440 @@ The larger or equal is intentional, because the mapping may already be not exhau
...
+because the extents and layout are part of the type.
+Even if all the extents are dynamic, the *rank* is static.
+
+If mdarray had a facility for extracting the container,
```
crtrott, line 836:
```suggestion
If `mdarray` had a facility for extracting the container,
```

P1684-mdarray/P1684.md
```diff
@@ -644,6 +651,440 @@ The larger or equal is intentional, because the mapping may already be not exhau
...
+then one way to implement an "out-of-place reshape"
+would be to extract the container from the source mdarray
```
crtrott, line 838:
```suggestion
would be to extract the container from the source `mdarray`
```

P1684-mdarray/P1684.md
```diff
@@ -644,6 +651,440 @@ The larger or equal is intentional, because the mapping may already be not exhau
...
+and move it into a new destination mdarray.
```
crtrott, line 839:
```suggestion
and move it into a new destination `mdarray`.
```

P1684-mdarray/P1684.md
```diff
@@ -644,6 +651,440 @@ The larger or equal is intentional, because the mapping may already be not exhau
...
+However, depending on the container type,
+this may copy the elements instead of viewing the original elements.
+An example of such a container type is `std::array`.
+
+Thus, the only way to implement a reshape operation
+that *always* views the original elements
+would be to use mdspan.  That is, the reshape operation
```
crtrott, line 846:
```suggestion
would be to use `mdspan`.  That is, the reshape operation
```

P1684-mdarray/P1684.md
```diff
@@ -644,6 +651,440 @@ The larger or equal is intentional, because the mapping may already be not exhau
...
+could view an existing mdarray as an mdspan
```
crtrott, line 847:
```suggestion
could view an existing `mdarray` as an `mdspan`
```

P1684-mdarray/P1684.md
```diff
@@ -644,6 +651,440 @@ The larger or equal is intentional, because the mapping may already be not exhau
...
+with different extents and/or layout.
+
+### Effects of extracting the container from an mdarray
+
+The discussion above explains why we do not think
+mdarray needs an operation for extracting its container.
+If we were nevertheless to add such an operation,
+this section explains its postconditions:
+that is, the state of the "extracted-from" mdarray.
```
crtrott, line 856:
```suggestion
that is, the state of the "extracted-from" `mdarray`.
```

P1684-mdarray/P1684.md
```diff
@@ -644,6 +651,440 @@ The larger or equal is intentional, because the mapping may already be not exhau
...
+
+If the interface is a member function, it must have return type `container_type&&` and be nonconst.  Making it rvalue reference qualified might incorrectly suggest that the layout mapping were moved-from.  In fact, the mdarray still has a perfectly valid layout mapping after the operation.
+
+#### Suggested wording
+
+```c++
+container_type&& extract_container();
+```
+
+*Effects*: Equivalent to: `return move(ctr_);`
```
crtrott, line 1042: We still might need both const and non const version if we do this:
```c++
const container_type&& extract_container() const;
```

Same effects.

crtrott (changes requested):
Like it, just some minor things.

P1684-mdarray/P1684.md
mhoemmen: What would it mean to extract the container from a const mdarray?

P1684-mdarray/P1684.md
mhoemmen: Fixed; thanks!

P1684-mdarray/P1684.md
mhoemmen: Would that imply that one could do this?

```c++
template<class E, class L, class C>
auto foo(const mdarray<E, L, C>& m) {
  return mdarray{m.extract_container(), 4, 6};
}

void bar() {
  const mdarray<float, layout_right> m0{2, 3, 4};
  mdarray m1 = foo(m0);
}
```

mhoemmen: @crtrott I don't think returning `const container_type&&` would make sense here.  The point of `extract_container()` is so that you can move the container out of the mdarray.  You can't move from a const rvalue.

P1684-mdarray/P1684.md
```diff
@@ -644,6 +651,482 @@ The larger or equal is intentional, because the mapping may already be not exhau
...
+
+int main()
+{
+  vector<float> storage{24, 0.0f};
+  pmr::monotonic_buffer_resource mbr{
+    storage.data(), sizeof(float) * storage.size()};
+  pmr::monotonic_buffer_resource* mbr_ptr = f1(f0(mbr));
+  // ... use mbr_ptr for other things ... 
+  return 0;
+}
```
dalg24, line 773: I have virtually no experience with polymorphic allocators but I don't think your code snippet does what you think.
It does not reuse the storage.  See https://godbolt.org/z/7accYa79z

P1684-mdarray/P1684.md
```diff
@@ -644,6 +651,482 @@ The larger or equal is intentional, because the mapping may already be not exhau
...
+be able to call functions that use a container
+with an incompatible allocator type.
+This is somewhat mitigated by `mdspan` being
+the preferred function parameter type for array views.
+Putting `mdarray` in the interface of a function
+expresses an ownership transfer.  That is,
+
+1. the only reason to expose `mdarray` as a function parameter
+    is if the caller is transferring ownership
+    of the `mdarray` to the function; and
```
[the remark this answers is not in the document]
mhoemmen: @dalg24 That's a good point, and I had thought about elaborating a bit here.  Here's what I might write instead.

There are two reasons to expose `mdarray` as a function parameter.

1. The caller is transferring ownership of the `mdarray` to the function.  This implies passing `mdarray` by rvalue reference.
2. The `mdarray` is small enough that passing by value is inexpensive (e.g., the object can fit in registers).

Otherwise, the correct idiom is to pass an `mdspan` that views the elements of the `mdarray`.

P1684-mdarray/P1684.md
mhoemmen: Thanks for pointing this out!  I didn't actually test this example.  I probably should have ; - ) .  It's been a couple years since I had to do anything with `std::pmr`.

P1684-mdarray/P1684.md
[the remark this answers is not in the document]
mhoemmen: Revised; thanks!

P1684-mdarray/P1684.md
```diff
@@ -644,6 +651,482 @@ The larger or equal is intentional, because the mapping may already be not exhau
...
+  // ... compute with m1 ...
+  return resource;
+}
+
+int main()
+{
+  vector<float> storage{24, 0.0f};
+  pmr::monotonic_buffer_resource mbr{
+    storage.data(), sizeof(float) * storage.size()};
+  pmr::monotonic_buffer_resource* mbr_ptr = f1(f0(mbr));
```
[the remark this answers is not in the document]
mhoemmen: Fixed; thanks!

P1684-mdarray/P1684.md
mhoemmen: I've fixed this example; thanks!

P1684-mdarray/P1684.md
```diff
@@ -644,6 +651,504 @@ The larger or equal is intentional, because the mapping may already be not exhau
...
+The `mdarray` would be unusable.
+Its destructor would still do the right thing,
+but users could no longer rely on
+any other properties of the `mdarray`.
+It would effectively be in a "valid but unspecified state,"
+just like its container.
+
+This means that generic code should treat
+an extracted-from `mdarray` like any moved-from object:
+the only safe operation is to let it be destroyed.
```
nliber, line 942: From a language point of view, I agree that objects need to be destructible without any preconditions.

From a library point of view, one can also assume that a type in a moved-from state can be assigned-to without any preconditions.  Is there a reason we don't say this?

And to be clear, neither of those is necessarily  a "valid but unspecified" state.  It can only be that if such a state maintains the class invariants, either directly or by adding preconditions on other operations.  This is sometimes referred to as a "partially formed state".

(Okay, it may not matter here since we are recommending not to have this functionality, but it does matter for the moved-from state of `mdarray`.)

nliber (approved):
I'm approving because the explanation is correct, even though I have reservations about not addressing the similar issue with a moved-from `mdarray`.

P1684-mdarray/P1684.md
```diff
@@ -644,6 +651,504 @@ The larger or equal is intentional, because the mapping may already be not exhau
...
+
+#### An "extracted-from" mdarray is like a moved-from object
+
+The hypothetical `extract_container()` member function
+would have the effect of returning `move(ctr_)`.
+If the caller uses the result in a `container_type`
+move constructor or move assignment operator,
+then this would leave the `mdarray`'s container in a moved-from state.
+Given that `mdarray` can accept any contiguous container
+and imposes no requirements on the container's moved-from state,
```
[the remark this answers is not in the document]
mhoemmen: @nliber This is a good point.  I really wish C++ had a "sink state" that the compiler could enforce, instead of whatever this moved-from state is supposed to be.  Here's a summary of my thoughts.

1. P1684 doesn't currently impose any requirements on the container's moved-from state, neither explicitly nor implicitly.  The move constructor and move assignment operator are both `= default`.  Thus, I have to conclude that a moved-from mdarray has a "valid but unspecified" container, a thing completely unusable for anything other than destruction.

2. Extraction is worse because it's not syntactically marked.  It's just some member function.  On the other hand, we could syntactically mark it by making it rvalue-qualified (`&&`).

3. If the container is `vector` or `vector`-like, it would be nice if returning a (named) mdarray from a function didn't double-allocate and copy.

P1684-mdarray/P1684.md
mhoemmen: @nliber Thanks for pointing this out!  

> From a library point of view, one can also assume that a type in a moved-from state can be assigned-to without any preconditions. Is there a reason we don't say this?

Is this part of the contiguous container requirements, or is this a general property of all C++ objects?  What in the Standard says that a moved-from object needs to be assignable?

nliber:
For all types specified _in the standard library_, there are only two possibilities.

1. Specify the moved-from state, such as we do with `vector`.
2. Say that the moved-from state is "valid but unspecified", which means we can call any function w/o a precondition on it, and we can call any function with a precondition as long as we can determine that precondition is satisfied.

I know of no types _in the standard library_ that have preconditions on destruction, or copy/move construction/assignment (for either source or destination).

In C++20 core made a change so that calling the pseudo-destructor on fundamental types essentially makes them unreadable until they are assigned-to.  (Unfortunately, this change inadvertently put a precondition on, say, copying a `vector<int>` which it didn't have before.   Library has not caught up with this.).

Looking at Stepanov/McJones Elements of Programming, they talk about the "partially formed state" where the only guaranteed operations are assign-to and destruction.

I believe that "partially formed state" is equavalent to being a specified state with weakened invariants which is also equivalent to adding preconditions on most functions stating that the object is not in that state.

That is essentially what the standard did with the `variant` `valueless_by_exception` state.  If you look at the specification, there are preconditions on most of functions involving `variant` that the `variant` isn't in that state.

The problem with all this comes because `mdarray` maintains an invariant over the container it owns.  If you don't know what the moved-from state is, you cannot know whether or not your invariants are still maintained.  If the underlying container is in a "valid but unspecifed" state, I believe we have to say what operations must be supported in the moved-from underlying container w/o needing a precondition to call them.  If we are going to do that (and I think we must), then destructibility and assign-to are the only reasonable ones, and we still have to specify everything else in terms of a precondition that the `mdarray` is not in the moved-from state (like we do in the `valueless_by_exception` state of `variant`, although in this case the state is not queryable).




If your class maintains an invariant over its members, then "valid but unspecified" is insufficient.

mhoemmen:
@nliber Thanks for explaining! : - )

> For all types specified in the standard library, ....

Alas, the cause of our concern is that we want to permit container types that are not in the standard library.  Perhaps that's just Not Worth It and we should stick with `vector` and `array`.

I agree that it would make sense to be able to assign to a moved-from mdarray, and that this reasonable thing calls for some requirements on the container.  (I was reminded today that `vector`'s move assignment operator is only conditionally `noexcept`.  What should mdarray do if an attempt to move from a `vector` throws?)

nliber:
@mhoemmen As long as we need to support `vector`, I don't see it being any more work making it a general adaptor.  Here is why:

The problem with an empty `vector` moved-from state is that it does not hold the invariant for an `mdarray` of all static extents, and there is no guaranteed way to put a `vector` in the correct state because when attempting so an exception may get thrown.

So we _still_ have to handle the case where `vector` is not in the correct state.  I see two approaches:

1. Say that source `mdarray` is in the equivalent of the "partially formed state", whether or not the move operation failed.  If the move operation failed, say that the destination `mdarray` is also in the equivalent of the "partially formed state".
2. Remove the move operations from `mdarray`.

Picking either of those is the same work whether we just support `vector` and `array` or if we keep it as a general adaptor.

mhoemmen:
@nliber Thanks! : - )  I would prefer Approach (1).  Removing the move operations from `mdarray` would make it prohibitively expensive to return a new mdarray from a function, for cases where copy elision is not required.

Do you have suggestions for wording?

mhoemmen:
> The problem with an empty `vector` moved-from state is that it does not hold the invariant for an mdarray of all static extents, ....

Just to clarify: I know that bit : - )  It sounds like we're just missing some wording.

nliber:
@mhoemmen I'll see what I can come up with.

PR #338, opened 2023-02-12
Remove note repeating what a note a few lines previous already said.

tzlaine:


mhoemmen (approved):
Thanks for the fix! : - )

This change has been implemented in PR https://github.com/ORNL/cpp-proposals-pub/pull/354 .

PR #339, opened 2023-02-12
Remove notes about how BLAS is organized, that its online reference c…

tzlaine:
…ontains a

typo, etc.

### DIFF
--- a/D1673/P1673.md
+++ b/D1673/P1673.md
@@ -6889,17 +6889,6 @@ or other things as appropriate.
 
 ### BLAS 1 functions [linalg.algs.blas1]
 
-<i>[Note:</i> The BLAS developed in three "levels": 1, 2, and 3.
-BLAS 1 includes vector-vector operations,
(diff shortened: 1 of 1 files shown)


D1673/P1673.md
```diff
@@ -6889,17 +6889,6 @@ or other things as appropriate.
 
 ### BLAS 1 functions [linalg.algs.blas1]
 
-<i>[Note:</i> The BLAS developed in three "levels": 1, 2, and 3.
```
mhoemmen, line 6892: @crtrott @dalg24 We might want to consider

* changing the section titles not to refer to BLAS 1, 2, or 3; and/or
* moving this Note to the front matter, so that readers have some context for the section titles.

Either option suggests a more elaborate change than just removing the note.  Would you concur?

mhoemmen:
I've addressed all but the one Note (see comment above) in https://github.com/ORNL/cpp-proposals-pub/pull/354 ; thanks!

PR #340, opened 2023-02-12
Add TODOs directly to text about a couple of notes that need fixing.

tzlaine:


D1673/P1673.md
```diff
@@ -6012,6 +6014,8 @@ constexpr auto scaled(
 [2]{.pnum} *Effects:* Equivalent to
 `return mdspan{x.data(), x.mapping(), accessor_scaled{alpha, x.accessor()}};`.
 
+(TZL) TODO: The non-collapsing behavior should be normative, and the rest of
```
mhoemmen, line 6017: @tzlaine Thanks for pointing this out!  Do you think the non-collapsing behavior would be sufficiently implied in the "Effects: Equivalent to" wording, or do we need more normative wording to require this?

tzlaine: Yes, it would -- good point.  "Equivalent to:" really means "exactly this code:".

D1673/P1673.md
```diff
@@ -5313,6 +5313,8 @@ access the upper triangle (`upper_triangular_t`) or lower triangle
 restrictions of `implicit_unit_diagonal_t` if that tag is also
 applied; see below.
 
+(TZL) TODO: I think you want the semantic part of this in normative wording.
```
mhoemmen, line 5316: Thanks for pointing this out, but I think the existing normative wording already conveys this.  Every algorithm takes an mdspan, and interprets the mdspan as an mdspan.  No normative wording says to treat an mdspan with `layout_transpose` differently than any other matrix.  In fact, most users won't ever encounter a `layout_transpose` mdspan (as it's possible to transpose any strided layout by swapping strides and extents).  For background, please refer to the following non-normative intro text.

* "[The BLAS is low level](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2023/p1673r11.html#blas-and-mdspan-are-low-level); it imposes no mathematical meaning on multidimensional arrays."

* https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2023/p1673r11.html#support-for-different-matrix-layouts (BLAS matrix "type" is not naturally represented by mdspan layout.  P1673 uses algorithms to distinguish matrix "type," and does not abuse layouts and accessors as "tags.")

* https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2023/p1673r11.html#interpretation-of-lower-upper-triangular (explaining why P1673's interpretation of Triangle differs from the BLAS's)

tzlaine: Ah, ok.  Perhaps it should just be stricken then?

mhoemmen: @tzlaine I never understood why the Standard doesn't like explanatory Notes ; - ) .  Yes, I think it should be struck.  I'm confident that the existing wording implies this interpretation of Triangle.  Any misunderstanding would come from people's experience with the (C or Fortran) BLAS, not with a straightforward reading of P1673.

Thanks! : - )

mhoemmen:
Changes implemented in PR https://github.com/ORNL/cpp-proposals-pub/pull/354 -- thanks!

PR #341, opened 2023-02-12
Moved preconditions that were listed in a couple of notes to proper

tzlaine:
precondition clauses.

### DIFF
--- a/D1673/P1673.md
+++ b/D1673/P1673.md
@@ -8922,10 +8922,7 @@ void matrix_product(ExecutionPolicy&& exec,
 #### Symmetric matrix-matrix product [linalg.algs.blas3.symm]
 
 <i>[Note:</i> These functions correspond to the BLAS function `xSYMM`.
-
-Unlike the symmetric rank-1 or rank-k update functions,
-these functions assume that the input matrix `A` --
-not the output matrix -- is symmetric. <i>-- end note]</i>
+<i>-- end note]</i>
 
 [1]{.pnum} The following requirements apply
 to all functions in this section.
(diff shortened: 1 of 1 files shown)


D1673/P1673.md
```diff
@@ -8938,6 +8935,8 @@ to all functions in this section.
 
   * [2.3]{.pnum} `C.extent(1)` equals `E.extent(1)` (if applicable).
 
+  * [2.4]{.pnum} The input matrix `A` is symmetric.
+
```
mhoemmen, line 8939: This is not actually correct.  The input matrix does not need to be symmetric.  The Remarks below explain that the function only accesses one triangle (lower or upper) of the matrix, and interprets the other triangle so that the algorithm effectively works with a symmetric matrix.

```suggestion
```

D1673/P1673.md
```diff
@@ -8922,10 +8922,7 @@ void matrix_product(ExecutionPolicy&& exec,
 #### Symmetric matrix-matrix product [linalg.algs.blas3.symm]
 
 <i>[Note:</i> These functions correspond to the BLAS function `xSYMM`.
-
-Unlike the symmetric rank-1 or rank-k update functions,
-these functions assume that the input matrix `A` --
-not the output matrix -- is symmetric. <i>-- end note]</i>
+<i>-- end note]</i>
```
mhoemmen, line 8925: Thank you for deleting this Note -- the Note was not actually correct, as my comment explains below.

mhoemmen (changes requested):
See comments -- we can just delete those lines from the Note here, without adding normative text.  We also need to look for and change any other places that call the matrix symmetric (or Hermitian) instead of relying on the Remarks.

D1673/P1673.md
```diff
@@ -9158,6 +9153,8 @@ not the output matrix -- is Hermitian. <i>-- end note]</i>
 
   * [2.3]{.pnum} `C.extent(1)` equals `E.extent(1)` (if applicable).
 
+  * [2.4]{.pnum} The input matrix is Hermitian.
```
mhoemmen, line 9156: Please see above note (the input matrix isn't Hermitian, it's interpreted as such; this should be covered by existing wording and changes in other PRs).

mhoemmen:
Changes implemented in PR https://github.com/ORNL/cpp-proposals-pub/pull/354 ; thanks!

PR #342, opened 2023-02-12
Remove reasoning about design decisions for notes.

tzlaine:


D1673/P1673.md
```diff
@@ -6568,10 +6525,6 @@ where
         then `add_const_t<ElementType>`;
 
     * [2.1.2]{.pnum} else, `ElementType`;
-        <i>[Note:</i> For general accessors,
```
mhoemmen, line 6571: We should _also_ remove this node because of feedback that we shouldn't try to const-ify the mdspan's `element_type`.  (It's not unreasonable to have a nonconst transposed view.  We might even want to consider removing the requirement that the result of `transposed` is read only.)

D1673/P1673.md
```diff
@@ -6600,10 +6553,7 @@ where
           `conditional_t<is_same_v<StorageOrder, column_major_t>, row_major_t, column_major_t>`;
 
     * [2.2.7]{.pnum} else if `Layout` is `layout_transpose<NestedLayout>`
-        for some `NestedLayout`, then `NestedLayout` <i>[Note:</i>
-        this optimizes applying `transposed` twice
-        to an `mdspan` with a custom layout,
-        as long as there are no intervening layout changes <i>-- end note]</i>;
+        for some `NestedLayout`, then `NestedLayout`
```
mhoemmen, line 6556:
```suggestion
        for some `NestedLayout`, then `NestedLayout`;
```

D1673/P1673.md
```diff
@@ -7004,19 +6932,7 @@ void givens_rotation_apply(
 ```
 
 <i>[Note:</i>
-These functions correspond to the BLAS function `xROT`.  `c` and `s`
-form a plane (Givens) rotation.  Users normally would compute `c` and
-`s` using `givens_rotation_setup`, but they are not required to do
-this.
-
-For `i` in the domains of both `x` and `y`,
```
mhoemmen, line 7012: We may want to consider making this math-font text normative.

mhoemmen (changes requested):
Thanks for doing this!  We can implement the one punctuation change.

D1673/P1673.md
mhoemmen: Actually, I think the Effects suffice here.

mhoemmen:
Changes made in PR https://github.com/ORNL/cpp-proposals-pub/pull/354 -- thanks!

PR #343, opened 2023-02-12
Remove implementation freedom notes.  I don't think these need to be

tzlaine:
explicitly covered in normative wording (but I may be wrong!).


mhoemmen (approved):
I agree; these are implementation freedom notes and thus can be removed.

I'm thinking about writing a separate nonnormative "commentary paper" that would include the text of these Notes.

mhoemmen:
These changes have been incorporated into PR https://github.com/ORNL/cpp-proposals-pub/pull/354 ; thanks!

PR #345, opened 2023-02-12
Replace the use of the word "method" with function; simplify "algorit…

tzlaine:
…hm and/or

function" text to just "function".  The word "method" has no meaning in the standard.

### DIFF
--- a/D1673/P1673.md
+++ b/D1673/P1673.md
@@ -5113,7 +5113,7 @@ void triangular_matrix_matrix_right_solve(
 [1]{.pnum} This clause lists the minimum requirements for all algorithms and classes in **[linalg]**,
 and for the following types:
 
-* [1.1]{.pnum} for any input or output `mdspan` parameter(s) of any algorithm or method in **[linalg]**,
+* [1.1]{.pnum} for any input or output `mdspan` parameter(s) of any function in **[linalg]**,
     the parameter(s)' `value_type` and `reference` type aliases;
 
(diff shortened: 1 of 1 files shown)


D1673/P1673.md
```diff
@@ -5113,7 +5113,7 @@ void triangular_matrix_matrix_right_solve(
 [1]{.pnum} This clause lists the minimum requirements for all algorithms and classes in **[linalg]**,
 and for the following types:
 
-* [1.1]{.pnum} for any input or output `mdspan` parameter(s) of any algorithm or method in **[linalg]**,
+* [1.1]{.pnum} for any input or output `mdspan` parameter(s) of any function in **[linalg]**,
```
mhoemmen, line 5116: @crtrott @dalg24 I think what I was getting at with "algorithm or method" is "function or class template."  The "class template" case is for the proxy reference types used by `conjugated` and `scaled`.

@tzlaine Would "function or class template" be a good way to express this?

tzlaine: I guess?  The way I read the previous text was "function or member function", which is why I rewrote it just to "function".  The following behavior that is described seems to apply to functions, their parameters, etc.  I did't see anything that looked like it pertained to properties of types per se.  If I missed that, then yes, "function or class"/"function template or class template" is the right thing to say.

mhoemmen:
The changes have been added to PR https://github.com/ORNL/cpp-proposals-pub/pull/354 -- thanks!

PR #346, opened 2023-02-12
When listing elements that apply to all functions in a certain part o…

tzlaine:
…f the

wording, call them elements, not requirements, since "requirements" is one of the kinds of elements.


mhoemmen (approved):
Good catch -- thank you!

mhoemmen:
Fixed in https://github.com/ORNL/cpp-proposals-pub/pull/354 ; thanks!

PR #347, opened 2023-02-12
Remove "we use `*` in a typename [as a wildcard]" language.  First, t…

tzlaine:
…here's no

"we" in the standard. :) Second, there do not appear to be any template parameters that contain "_2" (I did not search for the others), and I cannot find any template parameters that follow this sentence that contain a "*" either.  If there are, you shold just list all the variations, rather than including a pattern.


mhoemmen (approved):
We got rid of those `_1`, `_2`, etc. template parameters, so yes, this is outdated language -- good catch!

mhoemmen:
This change has been implemented in PR https://github.com/ORNL/cpp-proposals-pub/pull/354 -- thanks!

PR #348, opened 2023-02-12
Replace the use of "complex versions" of floating point types.  "complex

tzlaine:
versions" doesn't have any meaning in the standard.  To say what I think you meant, you should say "specializations of `complex`".  In the standard, a "specialization of `foo`" means an what we would verbally say is an "instantiation of the template foo".


mhoemmen (approved):
Thanks!

mhoemmen:
@tzlaine Thanks for the fix!  I've implemented this in PR https://github.com/ORNL/cpp-proposals-pub/pull/354 .  Once we have the changes from a PR in the draft, we'll close the PR, just to keep track of what we've done.

PR #349, opened 2023-02-12
In the standard, "Effects: Equivalent to: `X`" means, "The standard l…

tzlaine:
…ibrary

implementation mut write exactly `X` as the implementation of this function."

Also, "Effects: Equivalent to: `return X;`" is better writen as "Returns: X".

Finally, when listing multiple cases of what a function returns, it should look like this: https://eel.is/c++draft/containers#mdspan.layout.stride.obs .

IMPORTANT NOTE: This same pattern should be repeated for listing out the multiple cases of what a type might be.  For example, in 16.6.3, the part that says "Let `R` name the type ...", you wnat to do it like the return-cases example.  Something like:

Let `R` be `mdspan<ReturnElementType, Extents, Layout, ReturnAccessor>`, where:

  - `T` if some-condition.
  - Otherwise, `U` if some-other-condition.
  - ...
  - Otherwise, `Z`.

Also, as you can see there, you don't need to be so verbose as "names the type"; you can just say "is".

### DIFF
--- a/D1673/P1673.md
+++ b/D1673/P1673.md
@@ -5123,6 +5123,14 @@ and mathematical expression $EXPR$), the following apply:
     the result of the computation is written to the elements of the function
     parameter `R`.
 
+[2]{.pnum} For any function `F` that takes a `Triangle t` parameter, `t`
+applies to accesses to the previous parameter to the function.  Let `M` be
+such an access-modified function parameter.  `F` will only access the triangle
+of `M` specified by `t`.  For accesses `M[i, j]` in the domain of `M` but
+outside the triangle specified by `t`, `F` will interpret the value `M[i, j]`
+as equal to _`conj-if-needed`_`(M[j, i])` if the name of `F` starts with
+`hermitian`, and `M[j, i]` otherwise.
+
 ## Requirements [linalg.reqs]
 
 ### Value and reference requirements [linalg.reqs.val]
@@ -7818,12 +7826,6 @@ void matrix_vector_product(ExecutionPolicy&& exec,
 perform a count of `mdspan` array accesses and arithmetic operations
 that is linear in `x.extent(0)` times `A.extent(1)`.
 
-[6]{.pnum} *Remarks:* The functions will only access
-the triangle of `A` specified by the `Triangle` argument `t`,
-and will interpret the value `A[i,j]`
-as equal to `A[j,i]` for indices `i,j` in the domain of `A`
-but outside the triangle specified by `t`.
-
 ##### Overwriting symmetric matrix-vector product
 
 ```c++
@@ -7929,13 +7931,6 @@ void symmetric_matrix_vector_product(
 perform a count of `mdspan` array accesses and arithmetic operations
 that is linear in `x.extent(0)` times `A.extent(1)`.
 
-[6]{.pnum} *Remarks:* The functions will only access
-the triangle of `A` specified by the `Triangle` argument `t`,
-and will interpret the value `A[i,j]`
-as equal to _`conj-if-needed`_`(A[j,i])` for indices `i,j`
-in the domain of `A`
-but outside the triangle specified by `t`.
-
 ##### Overwriting Hermitian matrix-vector product
 
 ```c++
@@ -8037,16 +8032,10 @@ has the same type as the function's `Triangle` template argument.
 perform a count of `mdspan` array accesses and arithmetic operations
 that is linear in `x.extent(0)` times `A.extent(1)`.
 
-[6]{.pnum} *Remarks:*
-
-  * [6.1]{.pnum} The functions will only access the triangle of `A`
-      specified by the `Triangle` argument `t`.
-
-  * [6.2]{.pnum} If the `DiagonalStorage` template argument has type
-      `implicit_unit_diagonal_t`,
-      then the functions will not access the diagonal of `A`,
-      and will interpret `A` as if each of its diagonal elements
-      behaves as a two-sided multiplicative identity.
+[6]{.pnum} *Remarks:* If the `DiagonalStorage` template argument has type
+`implicit_unit_diagonal_t`, then the functions will not access the diagonal of
+`A`, and will interpret `A` as if each of its diagonal elements behaves as a
+two-sided multiplicative identity.
 
 ##### Overwriting triangular matrix-vector product [linalg.algs.blas2.trmv.ov]
 
@@ -8184,16 +8173,10 @@ the function's `Triangle` template argument.
 perform a count of `mdspan` array accesses and arithmetic operations
 that is linear in `x.extent(0)` times `A.extent(1)`.
 
-[6]{.pnum} *Remarks:*
-
-  * [6.1]{.pnum} The functions will only access the triangle of `A`
-      specified by the `Triangle` argument `t`.
-
-  * [6.2]{.pnum} If the `DiagonalStorage` template argument has type
-      `implicit_unit_diagonal_t`,
-      then the functions will not access the diagonal of `A`,
-      and will interpret `A` as if each of its diagonal elements
-      behaves as a two-sided multiplicative identity.
+[6]{.pnum} *Remarks:* If the `DiagonalStorage` template argument has type
+`implicit_unit_diagonal_t`, then the functions will not access the diagonal of
+`A`, and will interpret `A` as if each of its diagonal elements behaves as a
+two-sided multiplicative identity.
 
 ##### Not-in-place triangular solve [linalg.algs.blas2.trsv.not-in-place]
 
@@ -8523,11 +8506,6 @@ parameter, and `alpha` otherwise.  Computes $A = A + \alpha x x^T$.
 and arithmetic operations is linear in
 `x.extent(0)` times `x.extent(0)`.
 
-[6]{.pnum} *Remarks:* The functions will only access
-the triangle of `A` specified by the `Triangle` argument `t`,
-and will interpret the value `A[i,j]` as equal to `A[j,i]`
-for indices `i,j` in the domain `A` but outside that triangle.
-
 ##### Rank-1 update of a Hermitian matrix [linalg.algs.blas2.rank1.her]
 
 ```c++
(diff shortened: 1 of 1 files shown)


D1673/P1673.md
```diff
@@ -5106,6 +5106,34 @@ void triangular_matrix_matrix_right_solve(
...
+    the result of the computation is written to the elements of the function
+    parameter `R`.
+
+[2]{.pnum} For any function `F` that takes a `Triangle t` parameter, `t`
+applies to accesses to the previous parameter to the function.  Let `M` be
+such an access-modified function parameter.  `F` will only access the triangle
+of `M` specified by `t`.  For accesses `M[i, j]` in the domain of `M` but
+outside the triangle specified by `t`, `F` will interpret the value `M[i, j]`
+as equal to _`conj-if-needed`_`(M[j, i])` if the name of `F` starts with
+`hermitian`, and `M[j, i]` otherwise.
```
mhoemmen, line 5132: We might want to call out that we only need the `M[j, i]` branch for `symmetric_*` functions.  The `triangle_` functions actually assume that `M[i, j]` is zero outside the specified triangle.  (We might also want to consider adding that here.)

D1673/P1673.md
```diff
@@ -5106,6 +5106,34 @@ void triangular_matrix_matrix_right_solve(
...
+
+[3]{.pnum} Within all the functions in **[linalg]**, any calls to `abs`,
```
mhoemmen, line 5134: Excellent; this is much nicer!

D1673/P1673.md
```diff
@@ -6614,65 +6619,27 @@ where
...
-      for some `Triangle` and `StorageOrder`,
-      then equivalent to
-```c++
-return R(a.data(),
-  ReturnMapping(@_transpose-extents_@(a.mapping().extents())),
-  a.accessor());
-```
+  * [3.1]{.pnum} `R(a.data(), ReturnMapping(@_transpose-extents_@(a.mapping().extents())),
+      a.accessor())` if `Layout` is `layout_left`, `layout_right`, or a specialization of
+      `layout_blas_packed`.
```
mhoemmen, line 6626: I think breaking the line inside backquotes won't format correctly.  This also applies to the changes below.

```suggestion
  * [3.1]{.pnum} `R(a.data(), ReturnMapping(@_transpose-extents_@(a.mapping().extents())), a.accessor())`
      if `Layout` is `layout_left`, `layout_right`,
      or a specialization of `layout_blas_packed`.
```

D1673/P1673.md
```diff
@@ -7199,9 +7164,10 @@ T dot(ExecutionPolicy&& exec,
 `v2.static_extent(0)` equals `dynamic_extent`, then
 `v1.static_extent(0)` equals `v2.static_extent(0)`.
 
-[4]{.pnum} *Effects:* Let `N` be `v1.extent(0)`.
-If `N` is zero, returns `init`, else returns
-*GENERALIZED_SUM*(`plus<>()`, `init`, `v1[0]*v2[0]`, ..., `v1[N-1]*v2[N-1]`).
+[4]{.pnum} *Returns:*
+
+  * [4.1]{.pnum} `init` if `v1.extent(0) == 0`.
+  * [4.2]{.pnum} Otherwise, *GENERALIZED_SUM*(`plus<>()`, `init`, `v1[0]*v2[0]`, ..., `v1[N-1]*v2[N-1]`).
```
mhoemmen, line 7170: Do we need to define `N` here?

```suggestion
  * [4.2]{.pnum} Otherwise, *GENERALIZED_SUM*(`plus<>()`, `init`, `v1[0]*v2[0]`, ..., `v1[N-1]*v2[N-1]`), where `N` is `v1.extent(0)`.
```

D1673/P1673.md
```diff
@@ -7442,17 +7408,17 @@ one-norm for many linear algebra algorithms in practice. <i>-- end note]</i>
...
       `abs(real(v[0])) + abs(imag(v[0]))`, ...,
-      `abs(real(v[N-1])) + abs(imag(v[N-1]))`).
+      `abs(real(v[N-1])) + abs(imag(v[N-1]))`)
+      if `typename InVec::value_type`
+      is a specialization of `complex<R>`.
 
-  * [2.3]{.pnum} else, returns
+  * [2.3]{.pnum} Otherwise,
       *GENERALIZED_SUM*(`plus<>()`, `init`,
       `abs(v[0])`, ..., `abs(v[N-1])`).
```
mhoemmen, line 7423: Ditto, we need to define `N`.

D1673/P1673.md
```diff
@@ -7754,7 +7720,7 @@ void matrix_vector_product(ExecutionPolicy&& exec,
                            OutVec y);
 ```
 
-[1]{.pnum} *Effects:* Computes $y$ such that $y = A x$.
+[1]{.pnum} *Effects:* Computes $y = A x$.
```
mhoemmen, line 7723: Can we say this?  The issue is that in math font, the equals sign means equality.  There's no "assignment" in math.

mhoemmen (changes requested):
@tzlaine Thanks for all the suggestions!  @crtrott @dalg24 and I may need to take a closer look at this one.

D1673/P1673.md
tzlaine: Ah, I didn't get that subtlety.  Yeah, you should definitely add that as a third case.

D1673/P1673.md
tzlaine: Falcepalm!  I totally missed that `N` is re-used in the the `GENERALIZED_SUM`!

D1673/P1673.md
tzlaine: Right, but there's front matter that says how to read this.  First paragraph of [linalg.general] says that this means, "compute `$A x$`, and then assign it to the elements of function parameter `y`.  Any wording that uses math to describe the effects that is not this exact pattern (there are four instances of this) are worded a bit differently.

D1673/P1673.md
```diff
@@ -8082,16 +8035,10 @@ has the same type as the function's `Triangle` template argument.
...
-
-  * [6.1]{.pnum} The functions will only access the triangle of `A`
-      specified by the `Triangle` argument `t`.
-
-  * [6.2]{.pnum} If the `DiagonalStorage` template argument has type
-      `implicit_unit_diagonal_t`,
-      then the functions will not access the diagonal of `A`,
-      and will interpret `A` as if each of its diagonal elements
-      behaves as a two-sided multiplicative identity.
+[6]{.pnum} *Remarks:* If the `DiagonalStorage` template argument has type
```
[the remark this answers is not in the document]
tzlaine: +1

D1673/P1673.md
```diff
@@ -8563,26 +8502,13 @@ the function's `Triangle` template argument.
...
-
-  * [4.1]{.pnum} Overloads without an `alpha` parameter overwrite $A$
-      with $A'$ such that $A' = A + x x^T$,
-      where $x^T$ denotes the transpose of the column vector $x$.
-
-  * [4.2]{.pnum} Overloads with an `alpha` parameter overwrite $A$
-      with $A'$ such that $A' = A + \alpha x x^T$,
-      where the scalar $\alpha$ is `alpha`
-      and $x^T$ denotes the transpose of the column vector $x$.
+[4]{.pnum} *Effects:* Let $\alpha$ be $1$ for overloads without an `alpha`
```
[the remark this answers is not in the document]
tzlaine: Ah!  Very true.  Yeah, I think the old wording will have to suffice then.  I was trying to shorten.

D1673/P1673.md
```diff
@@ -10090,9 +9913,9 @@ nor `B.static_extent(0)`
 equals `dynamic_extent`, then
 `A.static_extent(1)` equals `B.static_extent(0)`.
 
-[5]{.pnum} *Effects:* Computes $X$ such that $AX = B$.
-If so such $X$ exists,
-then the elements of $X$ are valid but unspecified.
+[5]{.pnum} *Effects:* Computes $X'$ such that $AX' = B$, and assigns each
```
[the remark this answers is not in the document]
tzlaine: As I noted in an earlier comment above, the specific pattern `$x = EXPR$` has the meaning of computing `$EXPR$`, then assigning it to the function parameter `x`.  Here, we don't have that *exact* pattern, so I explained it as computing some intermediate result `$X'$`, and assigning *that* to `X`, element-wise.  This variance from the `$x = EXPR$` pattern only applies to four Effects: elements, as far as I can see.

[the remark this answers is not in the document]
mhoemmen: Thanks for explaining!  This pattern should apply to `triangular_matrix_vector_solve` (BLAS TRSV) as well as these BLAS TRSM functions.

D1673/P1673.md
```diff
@@ -6614,65 +6619,27 @@ where
...
-```c++
-return R(a.data(),
-  ReturnMapping(
-    @_transpose-extents_@(a.mapping().extents()),
-    array<decltype(), a.mapping().rank()>{
-      a.mapping().stride(1),
-      a.mapping().stride(0)}),
-    a.accessor());
-```
+  * [3.4]{.pnum} Otherwise, `R(a.data(), ReturnMapping(@_transpose-extents_@(a.mapping().extents()), array<decltype(), a.mapping().rank()>{a.mapping().stride(1), a.mapping().stride(0)}), a.accessor())` if `Layout` is `layout_stride`. (TZL) TODO: You have a nullary "decltype()"!
```
mhoemmen, line 6635: Nice catch! : - D  I'll just rely on CTAD for that `array` construction.

D1673/P1673.md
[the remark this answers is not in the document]
mhoemmen: Done in PR #354 !

D1673/P1673.md
```diff
@@ -8653,34 +8579,13 @@ the function's `Triangle` template argument.
...
-      overwrite $A$ with $A'$ such that $A' = A + x x^H$,
-      where $x^H$ denotes the conjugate transpose
-      of the column vector $x$.
-
-  * [4.2]{.pnum} Overloads with an `alpha` parameter
-      overwrite $A$ with $A'$ such that $A' = A + \alpha x x^H$,
-      where the scalar $\alpha$ is `alpha`
-      and $x^H$ denotes the conjugate transpose
-      of the column vector $x$.
+[4]{.pnum} *Effects:* Let $\alpha$ be $1$ for overloads without an `alpha`
```
mhoemmen, line 8582: (Please see above; thanks!)

D1673/P1673.md
mhoemmen: Done!

mhoemmen:
These changes have been implemented in PR #354 ; thanks!

PR #350, opened 2023-02-13
Fix a bunch of Preconditions: elements that should have been somethin…

tzlaine:
…g else

(usually Mandates:), or removed.

### DIFF
--- a/D1673/P1673.md
+++ b/D1673/P1673.md
@@ -6104,9 +6104,7 @@ constexpr explicit @_conjugated-scalar_@(Reference reference);
 constexpr static auto to_value(Reference value_type);
 ```
 
-[5]{.pnum} *Preconditions:* This function may be invoked arbitrarily many times.
-
-[6]{.pnum} *Effects:* Equivalent to
+[5]{.pnum} *Effects:* Equivalent to
 `return `_`conj-if-needed`_`(ReferenceValue(reference));`.
 
 ### Class template `accessor_conjugate` [linalg.conj.accessor_conjugate]
@@ -6342,10 +6340,9 @@ _`transpose-extents-t`_`<InputExtents>` names the `extents` type
       `OutputExtents::static_extent(r)`
       for 0 ≤ `r` < `InputExtents::rank()-2`.
 
-[3]{.pnum} *Preconditions:* `InputExtents`
-is a specialization of `extents`.
+[3]{.pnum} *Constraints:* `InputExtents::rank()` equals 2.
 
-[4]{.pnum} *Constraints:* `InputExtents::rank()` equals 2.
+[4]{.pnum} *Mandates:* `InputExtents` is a specialization of `extents`.
 
 ```c++
 template<class InputExtents>
@@ -6418,8 +6415,8 @@ public:
 };
 ```
 
-[7]{.pnum} *Preconditions:* `Layout` shall meet
-the `mdspan` layout mapping policy requirements.
+[7]{.pnum} `Layout` shall meet the `mdspan` layout mapping policy
+requirements.
 
 [8]{.pnum} *Constraints:* `Extents::rank()` equals 2.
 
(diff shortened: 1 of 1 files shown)


D1673/P1673.md
```diff
@@ -7318,11 +7315,12 @@ sum_of_squares_result<T> vector_sum_of_squares(
 <i>[Note:</i> These functions correspond to the LAPACK function `xLASSQ`.
 <i>-- end note]</i>
 
-[1]{.pnum} *Preconditions:*
+[1]{.pnum} *Mandates:*
 
-  * [1.1]{.pnum} `T` shall be *Cpp17MoveConstructible* and *Cpp17LessThanComparable*, and
+  * [1.1]{.pnum} `T` meets both the *Cpp17MoveConstructible* and
```
mhoemmen, line 7320: 1. Can this first point actually be a Mandate, given that the semantic parts of the named requirements aren't checkable at compile time?

2. Now that we have `std::move_constructible`, should we consider using that instead?  (LessThanComparable is a strict weak ordering, which doesn't have a named C++20 concept.)

D1673/P1673.md
```diff
@@ -7497,9 +7495,9 @@ typename InVec::size_type idx_abs_max(
 <i>[Note:</i> These functions correspond to the BLAS function
 `IxAMAX`. <i>-- end note]</i>
 
-[1]{.pnum} *Preconditions:* For `i` in the domain of `v`,
-let `abs_value_type` be `decltype(v[i])`;
-then, `abs_value_type` shall be *Cpp17LessThanComparable*.
+[1]{.pnum} *Mandates:* For `i` in the domain of `v`, let `abs_value_type` be
```
mhoemmen, line 7498: Same question as above.

mhoemmen (changes requested):

D1673/P1673.md
tzlaine: 1) Yes; this is done elsewhere in the standard already. 2) It depends on the exact semantics that you want, and I don't know the details well enough to answer that.  It might help to try it both ways and write some simple compile tests with different template parameters and see what makes the most sense.

D1673/P1673.md
```diff
@@ -6342,10 +6340,9 @@ _`transpose-extents-t`_`<InputExtents>` names the `extents` type
       `OutputExtents::static_extent(r)`
       for 0 ≤ `r` < `InputExtents::rank()-2`.
 
-[3]{.pnum} *Preconditions:* `InputExtents`
-is a specialization of `extents`.
+[3]{.pnum} *Constraints:* `InputExtents::rank()` equals 2.
```
mhoemmen, line 6343: It feels like the opposite would be easier: Mandate that the rank equals 2, and Constrain that `InputExtents` is a specialization of `extents`.

D1673/P1673.md
```diff
@@ -7318,11 +7315,12 @@ sum_of_squares_result<T> vector_sum_of_squares(
...
+[1]{.pnum} *Mandates:*
 
-  * [1.1]{.pnum} `T` shall be *Cpp17MoveConstructible* and *Cpp17LessThanComparable*, and
+  * [1.1]{.pnum} `T` meets both the *Cpp17MoveConstructible* and
+    *Cpp17LessThanComparable* requirements, and
 
-  * [1.2]{.pnum} `abs(x[0])` shall be convertible to `T`.
+  * [1.2]{.pnum} `abs(x[0])` is convertible to `T`.
 
 [2]{.pnum} *Constraints:* For all `i` in the domain of `v`,
```
mhoemmen, line 7325: This Constraint is left over from the "mathematical expression of the algorithm" R10 wording that R11 replaced.  I'll reword it.

mhoemmen: I've reworded this in PR https://github.com/ORNL/cpp-proposals-pub/pull/354 .

mhoemmen:
I've made these changes in PR https://github.com/ORNL/cpp-proposals-pub/pull/354 ; thanks!

PR #369, opened 2023-05-03
Added alternate constructor design to mdarray

nliber:
We discussed it over email a while ago.

Not sure if the alternate constructor design in the right place in the document.

Also, I didn't know what to do with the deduction guides, as the original ones are wrong for a variety of reasons:

* Guides 1 and 4 are marked `explicit`, but the corresponding constructor is implicit.
* Guides 2 and 3 are ambiguous with respect to each other, because `Mapping` and `Extents` are both unconstrained template parameters.  Same thing with guides 5 and 6, 8 and 9, 10 and 11.

### DIFF
--- a/P1684-mdarray/P1684.md
+++ b/P1684-mdarray/P1684.md
@@ -367,6 +367,13 @@ This is teachable with these simple rules:
 * Otherwise, if you have a container object, it comes first.
 * For constructors that take a parameter pack of extents (as opposed to an extents/mapping object), they come last.
 
+The deduction guides have been reworked for the new constructors.
+Because we need to deduce the container type, a `Container` parameter is
+provided for all the guides which are not using the default container,
+and that container is  then copied or moved into the `mdarray` on construction.
+This works because providing a container as the only parameter
+to the `Args...` parameter pack in any of the constructors effectively implements this.
+
 ::: cmptable
 #### Original (26 `mdarray` constructors)
 ```cpp
--- a/P1684-mdarray/P1684.md
+++ b/P1684-mdarray/P1684.md
@@ -1,6 +1,6 @@
 ---
 title: "`mdarray`: An Owning Multidimensional Array Analog of `mdspan`"
-document: P1684R4
+document: D1684R5a
 date: today
 audience: Library Evolution
 author:
@@ -14,10 +14,18 @@ author:
     email: <dansunderland@gmail.com>
   - name: Damien Lebrun-Grandie
     email: <lebrungrandt@ornl.gov>
+  - name: Nevin Liber
+    email: <nliber@anl.gov>
 ---
 
 
 # Revision History
+
+## P1684R4: pre-Varna mailing (2023-05)
+
+#### Changes from R4
+
+- Propose alternate constructor design design
 
 ## P1684R3: 2022-07 Mailing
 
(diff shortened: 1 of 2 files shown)


P1684-mdarray/P1684.md
```diff
@@ -334,6 +342,397 @@ In R3 we decided to radically cut back on constructors and simply leave those ou
 
 Finally, `mdarray` does not have the analog of `mdspan`'s constructor that takes an `array<IndexType, N>` of dynamic extents.  This avoids any ambiguity or confusion with a constructor that takes a container instance, in the case where the `container` happens to be an `array<IndexType, N>`.
 
+### Alternative constructor design
```
mhoemmen, line 345: Is this the "alternative" design (meaning the "Plan B" that we would accept in case LEWG rejected Plan A), or the design that we want LEWG to approve?  If the latter, should we consider not using the word "alternative" here?  For example, should we call this "New R5 constructor design"?

P1684-mdarray/P1684.md
```diff
@@ -334,6 +342,397 @@ In R3 we decided to radically cut back on constructors and simply leave those ou
...
+
+We can both significantly reduce the number of constructors *and*
+gain the ability to emplace the container.  This requires
+a bit of shuffling of the arguments, adding a tag and adding a parameter pack.
+
+* Whenever `extents_type` or `mapping_type` is provided, it must come first.
+* Whenever an `mdarray` is constructed with `val`, it must be preceded by a tag.
```
mhoemmen, line 352: Would you consider elaborating "`val`" just a bit?

```suggestion
* Whenever an `mdarray` is constructed with (the `value_type`) `val`, it must be preceded by a tag.
```

P1684-mdarray/P1684.md
```diff
@@ -334,6 +342,397 @@ In R3 we decided to radically cut back on constructors and simply leave those ou
...
+* Add a parameter pack `Args...` that essentially emplace constructs the container.
```
mhoemmen, line 353: The word "essentially" doesn't add meaning here.  Would you consider the following?

```suggestion
* Add a parameter pack `Args...` that is used to emplace construct the container.
```

P1684-mdarray/P1684.md
```diff
@@ -334,6 +342,397 @@ In R3 we decided to radically cut back on constructors and simply leave those ou
...
+
+The original adaptors couldn't do this, because variadics didn't exist in C++98.
```
mhoemmen, line 355: Would you consider elaborating "the original adapters" here?

```suggestion
The original container adaptors in the C++98 Standard Library couldn't do this, because variadic template parameters didn't exist in C++98.
```

P1684-mdarray/P1684.md
```diff
@@ -334,6 +342,397 @@ In R3 we decided to radically cut back on constructors and simply leave those ou
...
+We can and should do better now that we have better language facilities.
+
+In the original design,
```
mhoemmen, line 358: Does this mean the R4 mdarray design?  If so, would you consider clarifying that?

```suggestion
In the R4 mdarray design,
```

P1684-mdarray/P1684.md
```diff
@@ -334,6 +342,397 @@ In R3 we decided to radically cut back on constructors and simply leave those ou
...
+the unconstrained `Alloc` parameter is in a weird spot in the constructors
```
mhoemmen, line 359: Why is that "weird"?  Is the problem that this makes it impossible to add a parameter pack to be forwarded for emplace-constructing the container?

P1684-mdarray/P1684.md
```diff
@@ -334,6 +342,397 @@ In R3 we decided to radically cut back on constructors and simply leave those ou
...
+    constexpr mdarray(const mapping_type& m,
+                      const value_type& val,
+                      const Alloc& a);
+```
+
+```cpp
+  template<class... Args>
+    constexpr mdarray(const extents_type& ext,
+                      fill_t,
+                      const value_type& val,
```
mhoemmen, line 508: Should we consider adding an example to clarify why this is needed?  For instance, if `value_type` is `size_t`, just having the `Args&&...` wouldn't be enough to construct the container correctly.

P1684-mdarray/P1684.md
```diff
@@ -334,6 +342,397 @@ In R3 we decided to radically cut back on constructors and simply leave those ou
...
+             dextents<size_t, sizeof...(Integrals)>,
+             layout_right,
+             Container>;
+
+template<class Container, class Extents>
+mdarray(const Container&, const Extents&)
+  -> mdarray<typename Container::value_type,
+             Extents,
+             layout_right,
+             Container>;
```
mhoemmen, line 597: Do we think these deduction guides will work in practice?  Would we need to "constrain" `Extents` here?

P1684-mdarray/P1684.md
nliber: Well, it's the one **I** want LEWG to approve. :-)

I'll change the phrasing.  I want to emphasize that this hasn't been seen yet.

P1684-mdarray/P1684.md
nliber: Sure, no problem.

P1684-mdarray/P1684.md
nliber: Sure, no problem.

P1684-mdarray/P1684.md
nliber: I'll rephrase it as "the container adaptors introduced in C++98".  They aren't 100% the same because of move semantics (which leads to the current problem with `priority_queue` not guaranteed to maintain its invariants).

P1684-mdarray/P1684.md
nliber: Sure.

P1684-mdarray/P1684.md
nliber: It is a parameter that goes with the container, but is not right next to the container parameter, like it is for `stack`, `queue` and `priority_queue`.

P1684-mdarray/P1684.md
nliber: I think the `requires` clause at line 584 is sufficient to differentiate it from the other guides.

I don't know that it works in practice.  These two are identical CTAD constructors which map to two different types:

```c++
template<class Container, class Extents>
mdarray(Container&&, const Extents&)
  -> mdarray<typename Container::value_type, Extents, layout_right, Container>;

template<class Container, class Mapping>
mdarray(Container&&, const Mapping&)
  -> mdarray<typename Container::value_type, typename Mapping::extents_type, layout_right, Container>;
```

Another way to write the part before the `->` is:
```c++
template<class A, class B>
mdarray(A&&, const B&)
```

We'd be counting on `typename Mapping::extents_type` being SFINAEd out when `Mapping == std::extents<...>` AND being a better match when it is a `layout_*::mapping<...>`.  I don't think that works (but I could be mistaken).

mhoemmen: @nliber wrote:
> I don't know that it works in practice. 

Would you consider writing a brief test?

nliber: Let me try a simpler explanation.

Every function to the left of the `->` in the set of explicit deduction guides are functions added to the overload set when using `mdarray` without any angle brackets.  (There are also implicit deduction guides which also adds to the overload set.). Like every other kind of function in C++, one cannot have duplicates in the overload set.

Simple example:
```c++
template<typename T>
struct A {
    A(T&&) {}
};

template<typename T>
A(T&&)->A<T>;

template<typename T>
A(T&&)->A<T>;
```
[Compiler Explorer](https://godbolt.org/z/1Yah1KMaT) says:

```
<source>:10:1: error: deduction guide 'template<class T> A(T&&)-> A<T>' redeclared
   10 | A(T&&)->A<T>;
      | ^
<source>:7:1: note: 'template<class T> A(T&&)-> A<T>' previously declared here
    7 | A(T&&)->A<T>;
      | ^
Compiler returned: 1
```

mhoemmen: @nliber Would you consider adding this argument to the paper?

I'm hoping we can avoid the mdspan experience of trying to mash a bunch of CTAD fixes in near the end of the process : - ) .

crtrott:
Lets kill off the constructors taking container and then extents as a bunch of integers. That would mean you get:

```c++
mdarray()
mdarray(const mdarray&)
mdarray(mdarray&&)
template<..., class ... Args >
mdarray(const mdspan<...>&, Args&& ...)
template<..., class ... Args >
mdarray(const mdarray<...>&, Args&& ...)

template<class ... Exts >
mdarray(Exts ...)

template<class ... Args>
mdarray(const extent_type&, Args&& ... )
template<class ... Args>
mdarray(const mapping_type&, Args&& ... )

template<class ... Args>
mdarray(const extent_type&, fill_t, const value_type&, Args&& ... )
template<class ... Args>
mdarray(const mapping_type&, fill_t, const value_type&, Args&& ... )
```

crtrott:
Argument being, that other than for copy/converting constructors you always have the extents thingy first

crtrott:
I don't think this works.

Consider:
```c++
{
  vector<int> v(N);
  mdarray<int, dextents<int,1>> a(extents{N}, v);
}
{
  vector<int> v(N);
  mdarray<int, dextents<int,1>> a(extents{N}, std::move(v));
}
{
  vector<int> v(N);
  mdarray<int, dextents<int,1>> a(extents{N}, std::move(v), some_alloc);
}
```

How are these distinguishable?

crtrott:
Another complication:

```c++
// needs to call container_type(map.required_span_size(), some_alloc);
mdarray<int, dextents<int,1>> a(extents{N}, some_alloc);
// needs to call container_type(c, some_alloc);
mdarray<int, dextents<int,1>> a(extents{N}, c, some_alloc);
```

crtrott:
Also what about the 1D special case should that work?
```c++
mdarray<int, dextents<int,1>> a(c, some_alloc);
```

crtrott:
Ok, I just need to forward properly then it works.

PR #373, opened 2023-05-18
Batched BLAS white paper

mhoemmen:


batched_blas/batched_blas.md
```diff
@@ -16,13 +16,84 @@ toc: true
...
+
+"Batched" linear algebra functions solve
+many independent linear algebra problems all at once, in a single function call.  For example, a "batched GEMM" computes multiple matrix-matrix multiplies at once.
+Batched linear algebra interfaces have the following advantages.
+
+* By exposing the user's intent to solve many problems at once,
+    they expose much more potential parallelism and vectorization
+    than a single very small problem has (Dongarra 2018),
+    and amortize the overhead of representing each problem
+    as an argument of a function call.
```
kliegeois, line 33: The batched dimension is also an opportunity to improve the memory access pattern in the "Interleaved" and "Strided" cases described below.

mhoemmen: @kliegeois Would you consider writing something about that?

mhoemmen: @kliegeois Note that we don't have to repeat arguments already made in prior work by known experts.  We can just point readers to the relevant sections of those papers and tech reports.

kliegeois: @mhoemmen yes I can and will. I just need to think about something simple enough to explain this in couple of lines.

crtrott (commented):
Gonna merge this for now and do PR against it from now to change.

PR #375, opened 2023-05-19
Fix scaled(alpha, A) discussion in batched BLAS paper

mhoemmen:
@crtrott Per our discussion this afternoon, I've fixed the `scaled(alpha, A)` discussion in the batched BLAS paper.  It explains why `scaled` cannot take an array or mdspan of scaling factors, and therefore that the multiple-scaling-factors case needs overloads that take the scaling factors explicitly.

I've also removed the stub sections and added some links to existing vendor libraries.  The paper is ready for submission, though it could still use more work.


crtrott:
Did you look at my wording sketch? I think we should have that too in.

crtrott (approved):

PR #400, opened 2023-07-25
P1673 lwg review updated for 06-28 and 07-12

crtrott:
* make `scaled_accessor::reference` `const element_type`

* add converting, default and copy ctor to `scaled_accessor`

* rename `accessor_conjugate` to `conjugate_accessor`

* rename `accessor_scaled` to `scaled_accessor`

* fix example for `scaled`

* implement helper functions for algorithm mandates and preconditions
      and apply it to gemv

* fix hanging sections throughout

* use stable tags instead of "in this clause"

* make linalg-reqs-flpt a note

* fix linalg.algs.reqs constraint -> precondition

* fix unary plus and well formed


D1673/P1673.md
```diff
@@ -6496,29 +6519,34 @@ reference.  It is part of the implementation of `scaled`
 ```c++
 template<class ScalingFactor,
          class Accessor>
-class accessor_scaled {
+class scaled_accessor {
 public:
-  using reference = scaled_scalar<ScalingFactor,
-    typename Accessor::reference,
-    typename Accessor::element_type>;
   using element_type = add_const_t<typename reference::value_type>;
```
mhoemmen, line 6524: Looks like we need to update this line too.  (If `reference` is `element_type`, then `reference::value_type` likely doesn't exist or mean what we want.)

```suggestion
  using element_type = decltype(declval<ScalingFactor>() * declval<Accessor::element_type>());
```

D1673/P1673.md
```diff
@@ -6529,40 +6557,61 @@ private:
 
   * [2.3]{.pnum} `Accessor` meets the accessor policy requirements **[mdspan.accessor.reqmts]**.
 
+  * [2.4]{.pnum} _`scalable`_`<ScalingFactor, element_type>` is `true`.
```
mhoemmen, line 6560: I think this definition is circular, because the `element_type` of `scaled_accessor` necessarily comes from the product of the scaling factor and the nested accessor's element type.  Should we consider the following definition instead?

```suggestion
  * [2.4]{.pnum} _`scalable`_`<ScalingFactor, typename Accessor::element_type>` is `true`. 
```

D1673/P1673.md
```diff
@@ -6529,40 +6557,61 @@ private:
 
   * [2.3]{.pnum} `Accessor` meets the accessor policy requirements **[mdspan.accessor.reqmts]**.
 
+  * [2.4]{.pnum} _`scalable`_`<ScalingFactor, element_type>` is `true`. 
+
+  * [2.5]{.pnum} `is_constructible_v<element_type, decltype(declval<ScalingFactor>() * element_type)>` is `true`.
```
mhoemmen, line 6562: I'm not sure if we need this.  The definition of `scaled_accessor`'s `element_type` is "the result of the product of the scaling factor and the nested accessor's `element_type`."  If the product expression isn't well formed, then the definition of `element_type` isn't well-formed either, so we can't even talk about `element_type` as a requirement of `scaled_accessor`.

D1673/P1673.md
```diff
@@ -6529,40 +6557,61 @@ private:
...
+  * [4.1]{.pnum} Direct non-list-initializes _`scaling-factor`_ with `s`, and
+
+  * [4.2]{.pnum} direct non-list-initializes _`nested-accessor`_ with `a`.
 
 ```c++
 ```
 
-[4]{.pnum} *Returns:* `reference(`_`scaling-factor`_`,` _`nested-accessor`_`.access(p, i))`
+[5]{.pnum} *Returns:* `scaling_factor() * element_type(` _`nested-accessor`_`.access(p, i))`
```
mhoemmen, line 6595: This looks right.

D1673/P1673.md
```diff
@@ -6706,7 +6757,7 @@ public:
   using data_handle_type = typename Accessor::data_handle_type;
   using offset_policy = /* see below */;
 
-  constexpr accessor_conjugate(Accessor a);
+  constexpr conjugate_accessor(Accessor a);
 
   constexpr reference access(data_handle_type p, size_t i) const
```
mhoemmen, line 6762: This one is conditionally `noexcept`, but `scaled_accessor::access` is unconditionally `noexcept`.  I would favor the former.

D1673/P1673.md
```diff
@@ -6706,7 +6757,7 @@ public:
   using data_handle_type = typename Accessor::data_handle_type;
   using offset_policy = /* see below */;
 
-  constexpr accessor_conjugate(Accessor a);
+  constexpr conjugate_accessor(Accessor a);
```
mhoemmen, line 6760: Should we also add a converting constructor, like `scaled_accessor` has?

```c++
template<class OtherNestedAccessor>
  constexpr conjugate_accessor(const conjugate_accessor<OtherNestedAccessor>& other);
```

D1673/P1673.md
```diff
@@ -6529,40 +6557,61 @@ private:
...
+
 ```c++
-constexpr accessor_scaled(const ScalingFactor& s, const Accessor& a);
+template<class OtherScalingFactor, class OtherNestedAccessor>
+  constexpr scaled_accessor(const scaled_accessor<OtherScalingFactor, OtherNestedAccessor>& other);
 ```
 
+[3]{.pnum} *Constraints:*
+
+  * [3.1]{.pnum} `is_constructible_v<`_`nested-accessor`_`, const OtherNestedAccessor&>` is `true`, and
```
mhoemmen, line 6571:
```suggestion
  * [3.1]{.pnum} `is_constructible_v<`_`Accessor`_`, const OtherNestedAccessor&>` is `true`, and
```

D1673/P1673.md
mhoemmen: Actually, perhaps not -- I think the cast needs to use the nested accessor's element type.

```suggestion
[5]{.pnum} *Returns:* `scaling_factor() * typename Accessor::element_type(` _`nested-accessor`_`.access(p, i))`
```

D1673/P1673.md
```diff
@@ -6529,40 +6557,61 @@ private:
...
 
 ```c++
 constexpr ScalingFactor scaling_factor() const;
 ```
 
-[6]{.pnum} *Returns:* _`scaling-factor`_
+[7]{.pnum} *Returns:* _`scaling-factor`_
 
 ```c++
 constexpr Accessor nested_accessor() const;
```
mhoemmen, line 6611: `conjugate_accessor::nested_accessor()` returns `const Accessor&`.

D1673/P1673.md
crtrott: Oh this needed to be `using element_type = add_const_t<Accessor::element_type>;`

D1673/P1673.md
crtrott: we need to talk about this a bit more.

PR #401, opened 2023-08-02
Implement feedback from 07-26

crtrott:


D1673/P1673.md
```diff
@@ -713,6 +713,8 @@ toc: true
 
     * fix unary plus and well formed
 
+    * remove _`proxy-reference`_ , _`scaled_scalar`_ and _`conjugated_scalar`_
+
```
mhoemmen, line 717: I can't add a suggestion too far away from this, but should we also consider renaming `conjugate_accessor` to `conjugated_accessor`, as it better matches the function name `conjugated`?

D1673/P1673.md
```diff
@@ -3957,7 +3959,47 @@ on constraints wording.
...
+## Dummy Heading To Align Wording Numbering
+
+## Dummy Heading To Align Wording Numbering
+
+## Dummy Heading To Align Wording Numbering
+
+## Dummy Heading To Align Wording Numbering
+
+The preceding headings just push the automatic numbering of the document generator so that "Wording is in the place of 28.8 Numbers
```
mhoemmen, line 4000: It looks like there's a missing double quote after "Wording."

```suggestion
The preceding headings just push the automatic numbering of the document generator so that "Wording" is in the place of 28.8 Numbers
```

D1673/P1673.md
```diff
@@ -3981,17 +4023,16 @@ J. Demmel, I. Dumitriu, and O. Holtz,
 >
 > At the end of Table � ("Numerics library summary") in *[numerics.general]*, add the following: [linalg], Linear algebra, `<linalg>`.
 >
-> At the end of *[numerics]* (after subsection 28.8 *[numbers]*), add all the material that follows.
+> At the end of *[numerics]* (after subsection 28.8 *[numbers]*), add all the material that follows add the .
```
mhoemmen, line 4026: Not sure what happened here.  Did you perhaps mean to say "after the ..."?

D1673/P1673.md
```diff
@@ -4126,7 +4160,7 @@ concept @_out-object_@; = // exposition only
 template<class T>
 concept @_inout-object_@; = // exposition only
 
-// [linalg.algs], algorithms
+// [linalg.algs.blas1], BLAS-1 Functions
```
mhoemmen, line 4163: I do suppose they are "functions," but should we consider calling them "BLAS 1 algorithms," in order to emphasize how they are related to the existing Standard Algorithms?

```suggestion
// [linalg.algs.blas1], BLAS 1 algorithms
```

D1673/P1673.md
```diff
@@ -6521,8 +6343,8 @@ template<class ScalingFactor,
          class Accessor>
 class scaled_accessor {
```
mhoemmen, line 6344: I can't comment directly on the line, but should we consider making those unconditionally `noexcept` member functions conditionally `noexcept`?  For example, `access` would have `noexcept(noexcept(scaling_factor_* typename Accessor::element_type(accessor_.access(p, i))))`.

D1673/P1673.md
```diff
@@ -6542,22 +6364,22 @@ private:
   Accessor @_nested-accessor_@; // exposition only
 };
 
-template<class ScalingFactor, class ReferenceValue>
+template<class ScalingFactor, class ValueType>
 concept @_scalable_@ = // exposition only
 requires {
-  { declval<ScalingFactor>() * declval<ReferenceValue>() };
+  { declval<ScalingFactor>() * declval<ValueType>() };
```
mhoemmen, line 6370: Does `operator*` need to return non-`void` here?

D1673/P1673.md
```diff
@@ -6678,63 +6500,7 @@ If so, it could set the corresponding `TRANS*` BLAS function argument accordingl
...
-
-```c++
-constexpr static auto to_value(Reference value_type);
-```
-
-[5]{.pnum} *Returns:*
-_`conj-if-needed`_`(ReferenceValue(reference))`
-
-### Class template `conjugate_accessor` [linalg.conj.conjugate_accessor]
+#### Class template `conjugate_accessor` [linalg.conj.conjugate_accessor]
```
mhoemmen, line 6503: Should we consider calling this `conjugated_accessor`?

D1673/P1673.md
```diff
@@ -6750,12 +6516,16 @@ private:
...
+  using offset_policy = conjugate_accessor<Accessor::offset_policy>;
 
-  constexpr conjugate_accessor(Accessor a);
+  constexpr conjugate_accessor(const Accessor& acc);
+  constexpr conjugate_accessor() = default;
+  constexpr conjugate_accessor(const scaled_accessor&) = default;
+  template<class OtherScalingFactor, class OtherNestedAccessor>
+    constexpr conjugate_accessor(const conjugate_accessor<OtherNestedAccessor>& other);
 
   constexpr reference access(data_handle_type p, size_t i) const
```
mhoemmen, line 6530: This `access` function is conditionally `noexcept` (which is good).  Should we consider making the `access` function of `scaled_accessor` also conditionally `noexcept`?  Ditto for `offset`.

D1673/P1673.md
```diff
@@ -6764,82 +6534,59 @@ public:
...
-using offset_policy = /* see below */;
+template<class OtherScalingFactor, class OtherNestedAccessor>
+  constexpr conjugate_accessor(const conjugate_accessor<OtherScalingFactor, OtherNestedAccessor>& other);
 ```
 
-[6]{.pnum} This names
-
-* [6.1]{.pnum} `typename Accessor::offset_policy`,
-    if `remove_cv_t<typename Accessor::element_type>` is an arithmetic type;
+[4]{.pnum} *Constraints:* `is_constructible_v<`_`nested-accessor`_`, const OtherNestedAccessor&>` is `true`, and
```
mhoemmen, line 6562: I think we're missing the constraint for `OtherScalingFactor` here.  More importantly, _`nested-accessor`_ is a data member; we need to use a type there.

```suggestion
[4]{.pnum} *Constraints:*

  * [4.1]{.pnum} `is_constructible_v<Accessor, const OtherNestedAccessor&>` is `true`.

  * [4.2]{.pnum} `is_constructible_v<ScalingFactor, const OtherScalingFactor&>` is `true`.
```

This also suggests that it might be helpful to rename the template argument `Accessor` to `NestedAccessor`, so that the naming pattern is consistent.

D1673/P1673.md
```diff
@@ -6764,82 +6534,59 @@ public:
...
 
 ```c++
 constexpr reference access(data_handle_type p, size_t i) const
   noexcept(noexcept(reference(@_nested-accessor_@.access(p, i))));
 ```
 
-[8]{.pnum} *Returns:*
-`reference(`_`nested-accessor`_`.access(p, i))`
+[6]{.pnum} *Returns:*
+_`conj-if-needed`_`(`_`nested-accessor`_`.access(p, i))`
```
mhoemmen, line 6572: I think we may need to cast the result of the inner `access` call to `Accessor::element_type`.  The issue is that we define `element_type` as

```
using element_type = decltype(@_conj-if-needed_@(declval<Accessor::element_type>()));
```

Users could, in theory, provide an ADL-accessible `conj(Accessor::reference)` with a different return type than the ADL-accessible `conj(Accessor::element_type)`.  Casting the result to `Accessor::element_type` would make it unambiguous which `conj` is called.

Alternately, if `conj(Accessor::reference)` does lazy computation (e.g., deferring the expensive load of a proxy reference) while `conj(Accessor::element_type)` does eager computation, we might actually want `conj(Accessor::reference)` here.

In either case, we should make the definition of `element_type` consistent with the specification of the `access` function.

D1673/P1673.md
crtrott: Yeah sounds good

D1673/P1673.md
crtrott: yeah I think so.

D1673/P1673.md
crtrott: Yeah of course, I had that in mind but forgot.

D1673/P1673.md
```diff
@@ -6570,48 +6392,48 @@ template<class OtherScalingFactor, class OtherNestedAccessor>
...
 
-  * [4.2]{.pnum} direct non-list-initializes _`nested-accessor`_ with `a`.
+  * [5.2]{.pnum} direct non-list-initializes _`nested-accessor`_ with `a`.
 
 ```c++
 ```
 
-[5]{.pnum} *Returns:* `scaling_factor() * Accessor::element_type(` _`nested-accessor`_`.access(p, i))`
+[6]{.pnum} *Returns:* `scaling_factor() * Accessor::element_type(` _`nested-accessor`_`.access(p, i))`
```
mhoemmen, line 6415: I think this definition _is_ consistent with the definition of `element_type` above.  `conjugate(d)_scalar` has a consistency issue for its analogous `element_type` and `access`, though.

D1673/P1673.md
```diff
@@ -6570,48 +6392,48 @@ template<class OtherScalingFactor, class OtherNestedAccessor>
...
+[5]{.pnum} *Effects:*
 
-  * [4.1]{.pnum} Direct non-list-initializes _`scaling-factor`_ with `s`, and
+  * [5.1]{.pnum} Direct non-list-initializes _`scaling-factor`_ with `s`, and
 
-  * [4.2]{.pnum} direct non-list-initializes _`nested-accessor`_ with `a`.
+  * [5.2]{.pnum} direct non-list-initializes _`nested-accessor`_ with `a`.
 
 ```c++
```
mhoemmen, line 6412: Should we consider making this conditionally `noexcept`?

```suggestion
constexpr reference access(data_handle_type p, size_t i) const
  noexcept(noexcept(scaling_factor_* typename Accessor::element_type(accessor_.access(p, i))));
```

D1673/P1673.md
```diff
@@ -5303,7 +5341,7 @@ void triangular_matrix_product(
   InMat3 E,
   OutMat C);
 
-// [linalg.alg.blas3.rank-k.syrk],
+// [linalg.alg.blas3.rankk.syrk],
```
mhoemmen, line 5344: I do wish we could use underscores at least.

D1673/P1673.md
```diff
@@ -5670,65 +5708,83 @@ the "upper triangle," "lower triangle," and "diagonal" of a matrix.
...
 
+[*Example:*
+Small vector product accessing only specified triangle.
+It would not be a precondition violation for the non-accessed
+matrix element to be non-zero.
+
+```c++
+template<class Triangle>
+void triangular_matrix_vector_2x2_product(
+       mdspan<const float, extents<int, 2,2>> m,
```
mhoemmen, line 5736:
```suggestion
       mdspan<const float, extents<int, 2, 2>> m,
```

D1673/P1673.md
```diff
@@ -6147,57 +6199,17 @@ constexpr index_type stride(rank_type) const noexcept;
...
-<!--
-An implementation could dispatch to a function
-in the BLAS library, by noticing that the first argument
-has an `scaled_accessor` `Accessor` type.
-It could use this information to extract the appropriate
-run-time value(s) of the relevant BLAS function arguments
-(e.g., `ALPHA` and/or `BETA`),
-by calling `scaled_accessor::scaling_factor`.
--->
+### Exposition only helpers [linalg.helpers]
```
mhoemmen, line 6202:
```suggestion
### Exposition-only helpers [linalg.helpers]
```

D1673/P1673.md
```diff
@@ -6147,57 +6199,17 @@ constexpr index_type stride(rank_type) const noexcept;
...
-### Exposition-only function object _`abs-if-needed`_ [linalg.scaled.abs]
+#### _`abs-if-needed`_ [linalg.helpers.abs]
 
 [1]{.pnum} The name _`abs-if-needed`_ denotes an exposition-only function object.
 The expression _`abs-if-needed`_`(E)` for subexpression `E` whose type is `T`
 is expression-equivalent to:
 
-* [1.1]{.pnum} `E`
-    if `value_type` is an unsigned integer;
+* [1.1]{.pnum} `E` if `value_type` is an unsigned integer; otherwise
```
mhoemmen, line 6210: For consistency, should we consider attaching "otherwise" to the clause that explains what we do in that otherwise case?

```suggestion
* [1.1]{.pnum} `E` if `value_type` is an unsigned integer;
```

(and the change below)

D1673/P1673.md
```diff
@@ -6147,57 +6199,17 @@ constexpr index_type stride(rank_type) const noexcept;
...
 
-* [1.2]{.pnum} otherwise, `abs(E)`,
+* [1.2]{.pnum} `abs(E)`,
```
mhoemmen, line 6212:
```suggestion
* [1.2]{.pnum} otherwise, `abs(E)`,
```

mhoemmen (changes requested):

PR #413, opened 2023-10-12
P2897 (aligned_accessor): R1

### LINKED ISSUE #410: P2897R0 (aligned_accessor): LEWG feedback 2023/10/10
# LEWG review of P2897R0 (`aligned_accessor`) on 2023/10/10

## Feedback 

* Add example in the wording that uses the static member function to check pointer precondition
* Add example in the wording that uses `aligned_alloc` to create an overaligned allocation, to show that `aligned_accessor` exists as part of a system
* Change `gcd` converting constructor requirement from Constraint to Mandate
* Explore mdspan construction safety in a later paper (post-Kona)

Please see also https://github.com/cplusplus/papers/issues/1568 .

## Design alternatives

One design alternative for expressing the alignment precondition on the pointer would be to introduce an `aligned_pointer_t` type (not alias), as in the example currently in the mdspan repository.  The type's constructor would take a pointer and impose an alignment precondition on the pointer.  This would shift the burden for checking preconditions from the mdspan as a whole, to `data_handle_type` itself.  That way, mdspan itself wouldn't need a system for checking preconditions, yet (from the user's perspective) preconditions would still be checked at mdspan creation time.

mhoemmen:
# P2897R1, to be submitted 2023/10/15

## Make wording changes requested by LEWG review on 2023/10/10

- [x] Change `gcd` converting constructor Constraint to a Mandate
- [x] Add Example in the wording section that uses `is_sufficiently_aligned` to check the pointer overalignment precondition
- [x] Add Example in the wording section that uses `aligned_alloc` to create an overaligned allocation, to show that `aligned_accessor` exists as part of a system
- [x] Add an `explicit` constructor from `default_accessor`, so that users can type `aligned_mdspan y(x)` instead of `aligned_mdspan y(x.data_handle(), x.mapping(), aligned_accessor<element_type, byte_alignment>(x.accessor()))`.

## Add design discussion based on LEWG review on 2023/10/10

- [x] Explain why we do not include an `aligned_mdspan` alias

## Make other wording changes

- [x] Add to `aligned_accessor`'s Mandates that `byte_alignment >= alignof(ElementType)` is `true`.  This prevents construction of an invalid `aligned_accessor` object.

Fixes #410 .

### DIFF
--- a/aligned_accessor/P2897R1.html
+++ b/aligned_accessor/P2897R1.html
@@ -458,7 +458,9 @@ Purpose of this paper</a></li>
 <ul>
 <li><a href="#the-accessor-is-not-nestable" id="toc-the-accessor-is-not-nestable"><span class="toc-section-number">5.1</span> The accessor is not
 nestable</a></li>
-<li><a href="#we-do-not-define-an-alias-for-aligned-mdspan" id="toc-we-do-not-define-an-alias-for-aligned-mdspan"><span class="toc-section-number">5.2</span> We do not define an alias for
+<li><a href="#explicit-constructor-from-default_accessor" id="toc-explicit-constructor-from-default_accessor"><span class="toc-section-number">5.2</span> Explicit constructor from
+<code>default_accessor</code></a></li>
+<li><a href="#we-do-not-define-an-alias-for-aligned-mdspan" id="toc-we-do-not-define-an-alias-for-aligned-mdspan"><span class="toc-section-number">5.3</span> We do not define an alias for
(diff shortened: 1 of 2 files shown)


crtrott (commented):
I think this is ok. The examples are probably more than LWG would want.

mhoemmen:
@crtrott wrote:
>  The examples are probably more than LWG would want.

It's OK -- LEWG specifically asked for them.  We can always move them into the non-wording part of the paper later.

PR #435, opened 2023-11-10
[P2642] Remove constexpr/noexcept from defaulted special members

Quuxplusone:
constexprness/noexceptness are automatic for defaulted special members.
(That is, they're automatically _inferred with the correct values._ I have not confirmed whether all your `noexcept`s here were actually always-true; it's possible that by doing this I'm removing the possibility of "rogue `std::terminate`s," and you ought to consider that before merging.)


layout_padded/layout_padded.bs
dalg24, line 1: The layout padded spec is following what is already in the standard https://eel.is/c++draft/mdspan.layout.left.overview

Issue #444, opened 2024-02-07
P2642: Fixes for changes from LWG 2024/01/31 review


mhoemmen:
# Fixes for changes from LWG 2024/01/31

## [mdspan.layout.leftpadded.cons]

The `constexpr mapping(const extents_type& ext);` constructor currently has the following precondition (paragraph 2.2).

```
  * [2.2]{.pnum} If `extents_type::rank()` is greater than one
    and `padding_value` does not equal `dynamic_extent`,
    then $\mathit{least\_multiple\_at\_least}($ `padding_value` $,$ `ext.extent(0)` $)$
    is representable as a value of type `index_type`.
```

Given this wording, the following is not a precondition violation but will still produce offsets exceeding `MAX_INT`.

```c++
constexpr int e1 = MAX_INT / 3;
constexpr int s1 = 2*e1;
layout_left_padded<s1>::mapping<dextents<int, 2>> map(dextents<int, 2>{e1, 2}); 
```

@crtrott suggests the following wording.

```
  * [2.2]{.pnum} If `extents_type::rank()` is greater than one
    and `padding_value` does not equal `dynamic_extent`,
    then $\mathit{least\_multiple\_at\_least}($ `padding_value` $,$ `ext.extent(0)` $)$
    times the size of the multidimensional index space `ext` divided by `ext.extent(0)`
    is representable as a value of type `index_type`.
```

## Other

The Mandate "`padding_value` is representable as a value of type `index_type`" (without further qualifications) is wrong, because `dynamic_extent` is generally not representable as `index_type`.

`index_type @_stride-rm2_@ = @_static-padding-stride_@` is an unsafe conversion when _`static-padding-stride`_ is `dynamic_extent` and `index_type` is not `size_t`.   The default member initializer shouldn't take effect in that case, but the syntax itself might still be incorrect.

PR #445, opened 2024-02-09
Fix precondition

crtrott:


crtrott:
THis fixes the preconditions mentioned in #444

mhoemmen:
@crtrott This looks good!

Regarding zero padding value: The constructors that take a run-time padding value `pad` now explicitly forbid zero.  For (compile-time) `PaddingValue = 0` the definition of `least_multiple_at_least( padding_value , extents_type::static_extent(0) )` for `extents_type::static_extent(0)` not equal to `dynamic_extent` is now just `extents_type::static_extent(0)`, so I think this case is also fine now.

mhoemmen (approved):

PR #454, opened 2024-03-26
Paper that adds transposed special cases for P2642 layouts

### LINKED ISSUE #448: Add transposed special cases for P2642 layouts
# Add `transposed` special cases for P2642 layouts

## Problem

P1673's `transposed` doesn't have special cases for P2642's `layout_left_padded` and `layout_right_padded`.  As a result, `transposed` will return `layout_transpose` of these layouts, even though those layouts are more specialized than `layout_stride` and could thus easily have special cases.

P1673 used to have special cases for these two layouts, until the two layouts were moved from P1673 to P2642.

## Suggested fix

Add special cases to [[linalg.transp.transposed]](https://eel.is/c++draft/linalg#transp.transposed), paragraphs 3 and 4.

mhoemmen:
We propose to change the C++ Working Paper
so that `linalg::transposed` includes special cases
for `layout_left_padded` and `layout_right_padded`.
These are the two mdspan layouts proposed by P2642R6,
which was voted into the C++ Working Draft at the Tokyo 2024 WG21 meeting.
This change will make it easier for `linalg` implementations
to optimize for these two layouts by dispatching
to an existing optimized C or Fortran BLAS.
Delaying this until after C++26 would be a breaking change.

Fixes #448

### DIFF
--- a/transposed/transposed.md
+++ b/transposed/transposed.md
@@ -1,0 +1,251 @@
+
+---
+title: "Add transposed special cases for P2642 layouts"
+document: D????
+date: 2024/03/26
+audience: LEWG
+author:
+  - name: Mark Hoemmen
+    email: <mhoemmen@nvidia.com>
+toc: true
+---
+
+# Authors
+
+* Mark Hoemmen (mhoemmen@nvidia.com) (NVIDIA)
+
+# Revision history
+
+* Revision 0 to be submitted for the post-Tokyo mailing before 2024/04/16
+
+# Abstract
+
+We propose to change the C++ Working Paper
+so that `linalg::transposed` includes special cases
+for `layout_left_padded` and `layout_right_padded`.
+These are the two mdspan layouts proposed by P2642R6,
+which was voted into the C++ Working Draft at the Tokyo 2024 WG21 meeting.
+This change will make it easier for `linalg` implementations
+to optimize for these two layouts by dispatching
+to an existing optimized C or Fortran BLAS.
+Delaying this until after C++26 would be a breaking change.
+
+# Before and after comparison
+
+## Example
+
+The following example shows how this proposal
+changes the return type of `transposed`.
+
+```c++ 
+// optimized overload
+extern void some_algorithm(
+  mdspan<const float, dextents<size_t, 2>, layout_right_padded<dynamic_extent>> A_T,
+  mdspan<const float, dextents<size_t, 2>, layout_left_padded<dynamic_extent>> B,
+  mdspan<float, dextents<size_t, 2>, layout_left_padded<dynamic_extent>> C);
+
+template<class GenericFallBackLayout>
+void some_algorithm(
+  mdspan<const float, dextents<size_t, 2>, GenericFallBackLayout> A_T,
+  mdspan<const float, dextents<size_t, 2>, layout_left_padded<dynamic_extent>> B,
+  mdspan<float, dextents<size_t, 2>, layout_left_padded<dynamic_extent>> C)
+{
+  // ... slow generic code ...
+}
+
+void some_function(size_t N) {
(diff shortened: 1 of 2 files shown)


dalg24 (approved):
LGTM

transposed/transposed.md
```diff
@@ -0,0 +1,265 @@
...
+
+1. `decltype(A_00_T)::layout_type` is `layout_transposed<layout_left_padded<dynamic_extent>>`.
+
+2. Generic overload of `some_algorithm` is called.
+
+## After this proposal
+
+1. `decltype(A_00_T)::layout_type` is `layout_right_padded<dynamic_extent>`. 
+
+2. The statement `static_assert(A_00_T.stride(1) == 1);` is well formed.
```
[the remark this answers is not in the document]
mhoemmen: This is not wording, so I'm not too worried about it : - ) .  Thanks though!

transposed/transposed.md
```diff
@@ -0,0 +1,265 @@
...
+3. `submdspan_mapping` enables functionality --
+    the ability to slice `mdspan` with user-defined layouts --
+    while `transposed_mapping` would only enable (or simplify) optimizations.
+
+4. LEWG has already seen the proposed design over several reviews
+    (the last being the 2022/07/05 telecon review of P1673R9),
+    but has not yet had a chance to review
+    this alternative customization point design.
+
+We would like LEWG to poll this design alternative.
```
nmm0, line 229: Would it be possible to add `transposed_mapping` in the future even if it wasn't adopted for this paper?

transposed/transposed.md
```diff
@@ -0,0 +1,265 @@
...
+  // Approval of P2642R6 added this to the C++ Working Draft.
+  static_assert(is_same_v<
+    decltype(A_00)::layout_type,
+    layout_left_padded<dynamic_extent>>);
+  static_assert(A_00.stride(0) == 1); // compile-time value
+
+  mdspan A_00_T = linalg::transposed(A_00);
+  some_algorithm(A_00_T, B_00, C_00);
+}
+```
```
nmm0, line 84: I think the example is not easily-digestible. I would leave out B and C as they are not related to the choice of taking the fallback. I think it might be compelling to show a second mdspan with the transposed extents and basically just a comment next to `some_algorithm` calls for both mdspans showing that the second calls the optimized version while the first one doesn't because it's a transposed layout

transposed/transposed.md
mhoemmen: Hi @nmm0 !  It's possible in theory.  In practice, though, introducing `transposed_mapping` after C++26 would be a breaking change, as it would change the return type of `transposed` for user-defined layouts.

I personally prefer the `transposed_mapping` design, even if we only use it as a way to simplify wording.  However, WG21 may not like adding a customization point that will be rarely used, for a feature that is not strictly necessary in order to use or even to optimize std::linalg.

transposed/transposed.md
mhoemmen: @nmm0 Thanks for the feedback! : - )

It is a bit of a long example.  I mainly want to show how this proposal relates to P2642.

I could remove creation of separate `B_00` and `C_00` from `B` resp. `C` and just focus on `A_00`.  I would like to preserve the idea of calling an optimized algorithm, though.  What do you think?

transposed/transposed.md
mhoemmen: Thinking harder about `transposed_mapping` as a customization point made me change my mind; I'm now recommending against it.  I'll explain this in an updated draft.

transposed/transposed.md
mhoemmen: I've simplified the example a bit.  Please feel free to give more feedback if you like; thanks!

PR #437, opened 2023-11-14
Add P3050R0: "Optimize linalg::conjugated for noncomplex value types"

mhoemmen:
Abstract:

We propose the following change to the C++ Working Paper. If an `mdspan` object `x` has noncomplex `value_type`, and if that `mdspan` does not already have accessor type `conjugated_accessor<A>` for some nested accessor type `A`, then we propose to change `conjugated(x)` just to return `x`.

I submitted this already as P3050R0.

PR #455, opened 2024-04-08
Add P3050R1 ("Fix C++26 by optimizing linalg::conjugated for noncomplex value types")

mhoemmen:
Abstract:

We propose the following change to the C++ Working Paper.  If an `mdspan` object `x` has noncomplex `value_type`, and if that `mdspan` does not already have accessor type `conjugated_accessor<A>` for some nested accessor type `A`, then we propose to change `conjugated(x)` just to return `x`. Delaying this until after C++26 would be a breaking change.

I submitted this today (2024/04/08) as P3050R1.

This supersedes PR https://github.com/ORNL/cpp-proposals-pub/pull/437 , which I created in November 2023, and which still hasn't been merged.

### DIFF
--- a/conjugated/Makefile
+++ b/conjugated/Makefile
@@ -1,0 +1,3 @@
+include ../P0009/wg21/Makefile
+
+.DEFAULT_GOAL := $(HTML)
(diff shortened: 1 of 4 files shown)


mhoemmen:
Superseded by PR https://github.com/ORNL/cpp-proposals-pub/pull/470 .

PR #457, opened 2024-04-29
Added changes for LEWG review

crtrott:
- Split definition of _`atomic-ref-bound`_ into one for arithmetic and pointer, and one for other types
  - needed because the typedef `difference_type` should only appear for arithmetic and pointer types
- Added discussion for making _`atomic-ref-bound`_ exposition only or not
- Added discussion arguing for not having converting constructors with respect to memory order
- Added missing `fetch_min` and `fetch_max` functions
- Use "west const"
- Added header synopsis changes

### DIFF
--- a/atomic_accessor/atomic_accessor.md
+++ b/atomic_accessor/atomic_accessor.md
@@ -272,7 +272,7 @@ void compute_histogram(ExecT exec, float bin_size,
 }
 ```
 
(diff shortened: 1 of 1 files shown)


mhoemmen:
FYI, this version was submitted as P2689R3.

mhoemmen (approved):

PR #459, opened 2024-06-30
docs: pdf papers link

guspan-tanadi:


mhoemmen:
Thanks for the contribution!  Unfortunately, we aren't able to change P1673R13 any more, because it was the last submitted version.

PR #462, opened 2024-07-12
P3355R0: Add "Fix submdspan for C++26" proposal

mhoemmen:
Add P3355R0, "Fix submdspan for C++26."

We propose the following fixes to `submdspan` for C++26.

1. Permit user-defined pair types to be used as slices, rather than restricting the list to the opt-in set of types in **[tuple.like]**.

2. Change `layout_left`, `layout_right`, `layout_left_padded`, and `layout_right_padded` so that if a slice is `strided_slice` where `stride_type` models _`integral-constant-like`_ and `stride_type::value` equals 1, then the same layout mapping type results as if the slice models _`index-pair-like`_`<index_type>`.  This lets users express a slice with compile-time extent that does not cause common layouts to devolve into `layout_stride`.

### DIFF
--- a/submdspan-fixes/Makefile
+++ b/submdspan-fixes/Makefile
@@ -1,0 +1,3 @@
+include ../P0009/wg21/Makefile
+
+.DEFAULT_GOAL := $(HTML)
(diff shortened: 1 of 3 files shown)


mhoemmen:
Superseded by PR https://github.com/ORNL/cpp-proposals-pub/pull/474 .

PR #485, opened 2026-08-04
Start harmonizing atomic-ref-bounded with atomic_ref

crtrott:
- added store_op member functions
- added new floating point minimum and maximum member functions
- incorporated changes to value_type enabling a const T
- regrouped member functions

Note: I haven't done the constexpr everything thing yet. Also not added the new ctors yet or the address function.

### DIFF
--- a/atomic_accessor/atomic_accessor.md
+++ b/atomic_accessor/atomic_accessor.md
@@ -22,6 +22,14 @@ toc: true
 
 # Revision History
 
+## P2689R4 LEWG Feedback and catching up to `atomic_ref` changes
+
+- Harmonize with changes to `atomic_ref`
+  - added `store_op` member functions
+  - added new floating point minimum and maximum member functions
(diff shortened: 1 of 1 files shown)


atomic_accessor/atomic_accessor.md
mhoemmen, line 396:
```suggestion
constexpr bool @_atomic-ref-non-generic-type_@();
```

mhoemmen (approved):
Thanks Christian!  This is fine as-is.  I suggested one minor formatting change as a drive-by.  We'll need to go back through anyway to italicize and kabob-ize all the exposition-only things.

dalg24 (approved):
Fine

atomic_accessor/atomic_accessor.md
crtrott: Yeah there is actually quite a lot of things where we need to do the correct formatting for exposition only. I am planning on doing a PR just for that.

